PR #53, opened 2015-01-13
changes on dad decomplier module

nkbai:
delete outer classname when generate java source
for example :
	for class declaration:
	public class OuterClassName$InnerClassName --> public class InnerClassName
	for constructor:
	public OuterClassName$InnerClassName()-->public InnerClassName()


adesnos:
Sorry but we will not merge that. You could have duplicate class names.

PR #67, opened 2015-02-13
Modified indents in dvm.py

zyrikby:
Corrected indents in dvm.py (hope nothing has broken).


adesnos:
Sorry we will not merge that.

PR #68, opened 2015-02-13
Removed unused methods from dvm.py

zyrikby:
Removed duplicated method get_value(self) from EncodedArrayItem class
Removed unused method get_DVM_TOSTRING from DalvikVMFormat class


adesnos:
If it is unused it is fine so let's clean that later.

PR #85, opened 2015-04-01
fix some manifest issues

dweinstein:
This somewhat more closely aligns with the way [aapt](http://androidxref.com/4.4.3_r1.1/xref/frameworks/base/tools/aapt/Command.cpp#317) performs resolving of attributes and will help with apps that are using dexguard xml resource obfuscation.

On an obfuscated app one would not be able to correctly retrieve the `versionCode` or `versionName` fields from the manifest (and possibly many other fields).  It was discussed in this [stack overflow](http://stackoverflow.com/questions/26529120/how-can-i-obfuscate-xml-resource-files-with-dexguard)

> Each attribute in an XML resource file is identified by a name, but often also by a numeric identifier.  In AndroidManifest.xml for example:
android:versionName(0x010102lc) = "1.0"
In the obfuscation step, DexGuard can remove the name of an attribute, but only if this attribute also has a numeric identifier.
After DexGuard obfuscation the above attribute will look like this:
:(0x010102lc) = "1.0"


dweinstein:
I've tested a few apps and so far this works great but would appreciate some other eyeballs as well as more apps tested to ensure this doesn't break stuff :-)

subho007:
@dweinstein this is really good, but I guess instead of using ini files with all the constants and then reading it, can't it be put into one python file with a dictionary object ? Reading the whole file from the disk is always a costly process. What do you think?

dweinstein:
@subho007 thanks for the suggestion. Hopefully this is now more pythonic?

dweinstein:
Also FYI the values here are coming from the AOSP e.g.,  [`here`](http://androidxref.com/4.4.3_r1.1/xref/frameworks/base/core/res/res/values/public.xml)

subho007:
@dweinstein thank you so much for making the changes, I verified it also, this works perfectly :+1: :smiley:

adesnos:
Thx for the pull request !

PR #89, opened 2015-05-04
Adding api specific permissions

zyrikby:
This patch contains:
1. Tool to extract permission information from all manifest files in AOSP project and store the result in dictionary to describe permissions (in UTF-8 python module)
2. Modules with extracted permissions information for different versions of AOSP
3. Code to load api-specific modules
4. Small modifications to apk class to load permission and show them


subho007:
Wow, this is really an awesome PR. Let me verify this.

PR #90, opened 2015-05-05
Added get_NativeMethods and get_ReflectionCode

crozzy:


crozzy:
This pull request add the ability to use the output from `show_NativeMethods` and `show_RefectionMethods` by returning their results. 

There is some special formatting with the output of `show_RefectionMethods` that could be used, so avoid the temptation to just print the result of `get_ReflectionMethods`.

PR #93, opened 2015-05-06
The original readsleb128 has two major bugs.

fredrikluo:
- 1. result = unpack( '=b', buff.read(1) )[0]
     if result <= 0x7f:
        ...
     Wont work with multi-bytes leb128, as the "result" will be read
     as signed value and in multi-bytes, the MSB of all but the last
     byte are all set to 1.

     Therefore the "result", which is read as the first byte of a
     leb128 will always be negative, hence this condition test will
     always be true. Which means, it really just reads one byte and
     stops there for every signed leb128.

- 2. result = (result << 18) >> 18
     ...

     This , which I believe is to extend the sign bit, won't work with
     negative number neither, as after bit operator, result could become
     positive from negative.

Tese case like this:
t = -624485
assert (t == dvm.readsleb128_1(bytecode.BuffHandle(dvm.writesleb128(t))))

Would show what happens.

### DIFF
--- a/androguard/core/bytecodes/dvm.py
+++ b/androguard/core/bytecodes/dvm.py
@@ -214,7 +214,7 @@ def readsleb128(buff):
     result = 0
     shift = 0
 
-    while (True):
+    for x in range(0, 5):
        cur = ord( buff.read(1) )
        result |= (cur & 0x7f) << shift
        shift += 7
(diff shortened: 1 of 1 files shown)


fredrikluo:
I noticed that the DebugInfoItem is not used, and this bug could cause DebugInfoItem to incorrectly decode the debugging instructions.

PR #103, opened 2015-06-12
Permission api modification

zyrikby:
This patch adds support for showing the permissions declared within AndroidManifest file. Small modifications are also done with the names of the corresponding methods.


adesnos:
Hi,

please check your patch with the latest version of androguard. I removed few things related to the dynamic loading of modules which broke the session.

zyrikby:
Anthony, I resolved the conflict with the updates you've performed.

zyrikby:
There was a silly bug because I did not verify my code correctly with the last changes Anthony performed.

PR #105, opened 2015-06-17
Api permission mappings

zyrikby:
This PR provides a mapping between api calls and corresponding permissions depending on api version. The mappings are created using PScout tool. The PR also contains a tool that converts PScout mappings into the mappings used within AndroGuard.


adesnos:
Hi,

please check your patch with the latest version of androguard. I removed few things related to the dynamic loading of modules which broke the session.

PR #116, opened 2015-07-19
andrototal

nu11p0inter:
1. andrototal.py - prints results and (automatic) upload apk/dex files to virustotal
2. andrograph.py - file's autograph - show files signatures


adesnos:
Could you clean the second file ? It is a little messy.

nu11p0inter:
Definitely. 
Which file, and where...?

nu11p0inter:
The file is ready. clean & using androguard's code...

nu11p0inter:
The file is ready. clean & using androguard's code...

nu11p0inter:
Hi Antony,

I deleted the second file, using only the VT integration file.

It is not using VT-API because the API is limited for 4 requests per
minute. Since this is public code, we cannot use paid token, since it can
be easily abused by anyone.

So the parsing of the data must be manually... I performed validations so
in the case were their responses will change, it will notify the user to
request for update from Androguard.

Please let me know if you want to to add/ modify anything.

Best Regards,
Tal
http://lnkdin.me/cyber

PR #118, opened 2015-07-26
[core.androconf] fix name typo

leonid-s-usov:
Guys, there should be a better check of commits that are making it to the master ...


adesnos:
AOSP_PERMISSION_MAPPINGS doesn't exist.

PR #125, opened 2015-09-09
Apk resources

leonid-s-usov:
Although this is a working branch, it may be considered as a work in progress, mostly because of lack of testing.

The general problem was that androguard resource processing is not fully and sometimes not properly implemented. I tried to cover some important places where the lacking functionality was absolutely necessary (like proper StringBlock decoding and indirection support in resources). 

The resource config parsing is just partly covered. However, instead of putting more effort in completing the implementation I think that the right approach should be creating a bridge and linking to a native android code that works with resources. ([frameworks/base/libs/androidfw](https://android.googlesource.com/platform/frameworks/base/+/master/libs/androidfw/), [frameworks/base/tools/aapt](https://android.googlesource.com/platform/frameworks/base/+/master/tools/aapt/))


adesnos:
Yeah but we want only python code \o/

I will check the PR.

Thank you

leonid-s-usov:
Not sure why.. We can add the sources and setup tools will build it during install.  pyfuzzy dependency pulls quite a bunch of compiled sources

Sent from mobile device, typos and false corrections are possible

Regards,
Leonid S. Usov

9 вер. 2015 о 17:43 Anthony Desnos <notifications@github.com> написав(ла):

adesnos:
Yeah I will remove that !

adesnos:
Could you add your tests in tests/test_arsc.py please ?

leonid-s-usov:
Ok, I'll work on this over the weekend

Sent from mobile device, typos and false corrections are possible

Regards,
Leonid S. Usov

11 вер. 2015 о 13:18 Anthony Desnos <notifications@github.com> написав(ла):

leonid-s-usov:
@adesnos, re: tests. After looking and the test files, I don't quiet understand how it should work. They just dump the whole object, and return False. And that is a part of the AndroAuto functionality, so I can't just add more methods, I am bound to the `analysis_arsc`.. 
How would you like me to add the tests?

leonid-s-usov:
oh, probably you want me to merge your latest changes, right?

leonid-s-usov:
Does this look fine?

adesnos:
yep !

androguard/core/bytecodes/apk.py
```diff
@@ -770,6 +830,8 @@ def get_android_resources(self):
                 self.arsc["resources.arsc"] = ARSCParser(self.zip.read("resources.arsc"))
                 return self.arsc["resources.arsc"]
             except KeyError:
+                import traceback
```
adesnos: why do you add that ? Let's raise a new exception. Add a new class for that.

androguard/core/bytecodes/apk.py
```diff
@@ -913,75 +975,68 @@ def getString(self, idx):
...
-
-            self._cache[idx] = self.decode2(self.m_strings, offset, length)
+            self._cache[idx] = self.decode16(offset)
 
         return self._cache[idx]
 
     def getStyle(self, idx):
-        print idx
-        print idx in self.m_styleOffsets, self.m_styleOffsets[idx]
+        # FIXME
```
adesnos: remove the FIXME ?

adesnos:
lgtm, please fix the few hints.

androguard/core/bytecodes/apk.py
leonid-s-usov: New class seems too much in the case, I will just throw something appropriate

androguard/core/bytecodes/apk.py
leonid-s-usov: the problem was that I don't know how to check this function, and what it should have returned in the first place. Before it just had a number of printouts and returned nothing:) so FIXME here means "Please review this function someone who knows what it should do"

PR #134, opened 2015-09-27
Fixes #130 for unwanted characters

### LINKED ISSUE #130: IOError when running androdd.py
When running `androdd.py -i app-debug.apk -o tmp` on Windows I get following error:
```python
  File "E:\TestAndroid\androguard-2.0\androdd.py", line 218, in <module>
    main(options, arguments)
  File "E:\TestAndroid\androguard-2.0\androdd.py", line 203, in main
    export_apps_to_format(options.input, a, options.output, options.limit, options.jar, options.decompiler, options.format)
  File "E:\TestAndroid\androguard-2.0\androdd.py", line 195, in export_apps_to_format
    with open(filename + ".ag", "w") as fd:
IOError: [Errno 22] invalid mode ('w') or filename: 'tmp/android/support/v4/BuildConfig/<init>--V.ag'
```

subho007:


F3real:
Same thing should be done in pydot.py because error occurs there too if u specify format in androdd.py. 
```
 line 1809, in <lambda> lambda path, f=frmt, prog=self.prog : self.write(path, format=f, prog=prog))
 line 1895, in write dot_fd = file(path, "w+b")
```

subho007:
@F3real have you checked it with the following patch ? Pydot is a library, one should pass in the correct filename to the library.

F3real:
I downloaded  whole androdd.py from pull request and ran it and same thing happened. I got error with <init>--V.png file. I only added import from ```androguard.core.androgen import Androguard``` because it showed me error otherwise. You can try yourself with -f png flag and see if it generates files with bad names.

F3real:
Btw I am not really good with Python but I think you should add import for string library and change 
```filename = ''.join(c for c in s if c in valid_chars)``` to
```filename = ''.join(c for c in filename if c in valid_chars)```

subho007:
@F3real ahh good catch, during debugging, I use short variable names. Can you re-check with my branch once again ? This should work.

subho007:
@F3real I don't have a windows machine to check

F3real:
I am not really good with github, so I don't know how to do that.  I applied changes I said earlier and it works for me now. Only bug it had was because $ was missing from valid chars so I was getting errors for inner classes.

subho007:
@F3real can you run these commands to check?

`git reset --hard HEAD `
`git pull origin pull/134/head `

F3real:
fatal: Couldn't find remote ref pull/134/head
Unexpected end of command stream

PR #142, opened 2015-10-18
Replace distribute dependency with setuptools

thedrow:
Also, it's no use to require it during the installation process. Instead, specify that it is required for setup.


delijati:
Could this be merged `distribute` is deprecated.

subho007:
@reox it seems you have already fixed it in your PR. Can we close this?

PR #143, opened 2015-10-19
Fix APK signature methods.

strazzere:
APKs can be signed using RSA, DSA and EC keys. This
change allows you to be able to see the EC key suffix
when searching an APK. I've also added new methods to
return a list of _all_ signatures found opposed to just
the first.

The older methods are still available for backwards
compatability. They have been refactored to select
the first of the list, which is how the previous
functionality would have worked.


adesnos:
Thx Tim !

PR #144, opened 2015-10-22
added SIGNATURE definitions

oguzhantopgul:
added SIGNATURE definitions that were deleted at commit
6b24ddf7d385e8a9062de7ae95d7e8dfb12f7b1c which results an error when
androsign, androcsign and andrdiff tolls are run.


adesnos:
Sorry but I removed all these code right now, it is unstable, use the release tar.gz

PR #146, opened 2015-10-27
apk: build locale code properly

ganadist:
Signed-off-by: Young-Ho Cha ganadist@gmail.com

### DIFF
--- a/androguard/core/bytecodes/apk.py
+++ b/androguard/core/bytecodes/apk.py
@@ -1767,17 +1767,14 @@ class ARSCParser(object):
                     if header.type == RES_TABLE_TYPE_TYPE:
                         a_res_type = self.packages[package_name][nb + 1]
 
-                        if a_res_type.config.get_language(
-                        ) not in self.values[package_name]:
-                            self.values[package_name][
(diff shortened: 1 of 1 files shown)


reox:
Is there a specific reason to concatenate the locales like this: `locale = "-r".join((language, region))`?

In my opinion the result should be `"{}_{}".format(language, region)` or if you follow the structure that is given in the documentation: `"-b+{}+{}".format(language, region)`

ganadist:
> In my opinion the result should be "{}_{}".format(language, region)
It means "{}_r{}".format(language, region) ?

reox:
yes, but for what is the `r`? In my opinion a locale code is for example `en_US` and not `en_rUS`?

ganadist:
In https://developer.android.com/guide/topics/resources/providing-resources.html#table2 , describes about locale.

> The language is defined by a two-letter ISO 639-1 language code, optionally followed by a two letter ISO 3166-1-alpha-2 region code (preceded by lowercase "r").

reox:
Ahh I see, thank you! Now I understand

Issue #148, opened 2015-11-18
getting activity list or any elements list fails (unicode error)


ranjeetDT:
def get_elements(self, tag_name, attribute):
        """
            Return elements in xml files which match with the tag name and the specific attribute

```
        :param tag_name: a string which specify the tag name
        :param attribute: a string which specify the attribute
    """
    l = []
    for i in self.xml:
        for item in self.xml[i].getElementsByTagName(tag_name):
            value = item.getAttributeNS(NS_ANDROID_URI, attribute)
            value = self.format_value(value).encode('ascii', 'ignore') <-- this line should be added
            l.append(str(value))
    return l
```

test 'callsign android app' , https://play.google.com/store/apps/details?id=com.callsign.android.app

error is 

File "lib/androguard/core/bytecodes/apk.py", line 647, in get_activities
    return self.get_elements("activity", "name")
  File "lib/androguard/core/bytecodes/apk.py", line 569, in get_elements
    value = self.format_value(value).decode('ascii', 'ignore')
UnicodeEncodeError: 'ascii' codec can't encode character u'\u0493' in position 2: ordinal not in range(128)

ThomasBordenave:
I have the same error when using API get_services(). Of course, get_elements() is called.

```bash
Traceback (most recent call last):
    File "test.py", line 53, in <module>
        for s in a.get_services():

    File "build/bdist.linux-x86_64/egg/androguard/core/bytecodes/apk.py", line 510, in get_services
    File "build/bdist.linux-x86_64/egg/androguard/core/bytecodes/apk.py", line 436, in get_elements
UnicodeEncodeError: 'ascii' codec can't encode characters in position 2-3: ordinal not in range(128)
```

leonid-s-usov:
Not sure which version of the code you are using. I know that master is broken as many report, but it contains a my pull request merged, where I have fixed a big mess in the encoding when reading string blocks from packed files. Try patching your version with this https://github.com/leonid-s-usov/androguard/commit/ab2fbac1b11002581a454048417a3930c7d88b32

ThomasBordenave:
Thanks for your answer, I'm using the last stable version. I have almost the same issue with this patch:

```bash
Traceback (most recent call last):
  File "test.py", line 53, in <module>
    for s in a.get_services():
  File "build/bdist.linux-x86_64/egg/androguard/core/bytecodes/apk.py", line 572, in get_services
[...]

PR #151, opened 2015-11-26
Correct encoding for non-ascii values

ThomasBordenave:
Non-ASCII characters are found in the list of services (call of get_services() function).

Fix bug found with .apk with sha256:
0b8ba0c6cebe5695639bf1b282b52f126dba733f3c204e37615a3ba5f7dd6fe8

See issue #148 for more details.

### DIFF
--- a/androguard/core/bytecodes/apk.py
+++ b/androguard/core/bytecodes/apk.py
@@ -527,7 +527,7 @@ class APK(object):
                 value = item.getAttributeNS(NS_ANDROID_URI, attribute)
                 value = self.format_value(value)
 
-                l.append(str(value))
+                l.append(unicode(value))
         return l
 
     def format_value(self, value):


leonid-s-usov:
is this the only place where str() is used on the values from XML? I think there are more places like that, maybe it is appropriate to fix all of them at once

ThomasBordenave:
Ok, I'll take a look a the code.

ThomasBordenave:
After a quick look, I don't see anymore XML parsing but I might be wrong, I don't really have time today.

leonid-s-usov:
The only place I have found in that file is [in the __init__ method](https://github.com/ThomasBordenave/androguard/blob/aebd2cd91842a576b2afdb4afa3560563b9d5172/androguard/core/bytecodes/apk.py#L215) where permissions are read out. Not sure if all of the fields need a unicode conversion due to the "identifier" nature of those, but there is a `description` field.. 

Even more than that, after thinking about it for a while, I started doubting if _any_ conversion is needed in all places where xml attributes are being read. Why is it there in the first place? shouldn't we just remove all `str()` around values read from the XML?

PR #153, opened 2015-11-30
Fix 309aa68 to properly check if APK.

strazzere:
This is actually now similar to how TIKA checks if it
is an APK by checking if;
1. Magic is PK (zip)
2. Contains a Android Manifest (required)
3. Is signed (Manifest.mf)


subho007:
@reox I guess we can close this PR since you already implemented this and merged it ?

PR #155, opened 2015-12-10
Arsc corrections

leonid-s-usov:
The only APK so far that can be used to reproduce a crash which is fixed by this PR can be downloaded [from google play](https://play.google.com/store/apps/details?id=com.alibaba.intl.android.apps.poseidon)

I am not sure if it is OK to include 3rd party APK or even it's resources file into the test suite, so I didn't do that.

On the other hand, I am also not aware of an easy way to generate an arsc file with the non-standard header sizes (which is the reason for the crash).

### DIFF
--- a/androguard/core/bytecodes/apk.py
+++ b/androguard/core/bytecodes/apk.py
@@ -965,18 +965,13 @@ def show_Certificate(cert):
 # http://code.google.com/p/android4me/source/browse/src/android/content/res/AXmlResourceParser.java
 
 UTF8_FLAG = 0x00000100
-CHUNK_STRINGPOOL_TYPE = 0x001C0001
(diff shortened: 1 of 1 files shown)


subho007:
@leonid-s-usov can you retest this PR against the latest code we got in our master?

leonid-s-usov:
I have just looked as the master source code of the ARSCParser and it looks like the fixes from the PR are still needed. 
I havent been looking into this for 2 years already but I do remember that these changes are absolutely required if you want to get a parser compatible with the specification of the arsc format. I have worked with the original android source code to create this pull request in order to correctly process resource files.

There are just 2 trivial conflicts. I think you will benefit from getting this one merged

PR #179, opened 2016-05-18
Improvements to dvm.py

erezsh:
- Removed unnecessary string processing
- Turned returned strings to be more smali-compliant
- Added several utility methods


totoag:
obsolete probably :)

erezsh:
One can only hope

PR #184, opened 2016-06-07
Fix getAttributeNS may be return unicode and can't transform return v…

billvsme:
getAttributeNS may be return unicode string.
like this app: http://f1.market.mi-img.com/download/AppStore/084a35f702935823c9b2f0518776392209b40ad4a/com.tencent.mobileqq.apk

may be raise:
     UnicodeEncodeError: 'ascii' codec can't encode characters in position 2-5: ordinal not in range(128)


reox:
@billvsme i checked your file with the current master and it seems to work without your PR. Is it still an issue?

subho007:
these are handled automatically by python3. This issue is resolved. closing this

PR #196, opened 2016-10-05
Script that outputs Dalvik byte code of APK in a file for analysis

mohsinjuni:
I have used this modified script to generate Dalvik byte code for analysis. We built a tool to analyze the output generated from this script. Thought it may be helpful to others as well.

### DIFF
--- a/androguard/core/analysis/analysis.py
+++ b/androguard/core/analysis/analysis.py
@@ -280,10 +280,8 @@ class BasicBlocks(object):
         This class represents all basic blocks of a method
     """
 
-    def __init__(self, _vm, tv):
+    def __init__(self, _vm):
         self.__vm = _vm
-        self.tainted = tv
-
         self.bb = []
 
     def push(self, bb):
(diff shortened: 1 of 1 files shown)


reox:
there are probably a ton of useful scripts out there. The problem is, some work only with a specific version; others require some libraries.
I think it is a good idea to keep them.

In this case: there are some imports that does not work anymore.
@mohsinjuni can you check if this script still works with the current master?

for the general case: how can we mark that certain scripts/tools only work with a specific version? Should old scripts kept or upgraded? (See also the demo folder...)

reox:
I just saw that you duplicated a lot of functions in there. is there any reason to do this?

mohsinjuni:
There is no reason for duplicating code. I had copied one of the existing scripts and modified it for my purpose. I just tried this script with master version of Androguard and it says, "no module named androgen". However, this script still works with my older version of Androguard.

subho007:
This script will no longer work on the newer androguard. Please send in a new PR with the newer version of Androguard.

Issue #124, opened 2015-09-03
libmagic broken on mac


nachtmaar:
After installing libmagic with

```
brew install libmagic
```

androguard fails with the following traceback:

```
RROR: 'module' object has no attribute 'open'
Traceback (most recent call last):
  File "/Users/nils/uni/android_security/workspace/androlyze_stuff/androlyze/androlyze/analyze/AnalyzeUtil.py", line 64, in open_apk
    eandro_apk = EAndroApk(apk_or_path)
  File "/Users/nils/uni/android_security/workspace/androlyze_stuff/androlyze/androlyze/model/android/apk/EAndroApk.py", line 29, in __init__
    APK.__init__(self, *args, **kwargs)
  File "/Users/nils/uni/android_security/workspace/androlyze_stuff/androlyze/androguard/androguard/core/bytecodes/apk.py", line 222, in __init__
    self.get_files_types()
  File "/Users/nils/uni/android_security/workspace/androlyze_stuff/androlyze/androguard/androguard/core/bytecodes/apk.py", line 339, in get_files_types
    ms = magic.open(magic.MAGIC_NONE)
AttributeError: 'module' object has no attribute 'open'
```

nachtmaar:
The following code works on MAC:

```
import magic
with magic.Magic() as m:
        print.id_buffer(raw_file)
       # or: m.id_filename(file_path)

```

subho007:
Works good on my mac without the code changes you suggested.

subho007:
@nachtmaar as specified in the code at https://github.com/androguard/androguard/blob/master/androguard/core/bytecodes/apk.py#L392

You need to install python-magic and not libmagic

to install python-magic you can use:

`pip install python-magic`

lucasduffey:
Using python2 on ubuntu 16 LTS, you want to get python-magic from apt (not from pip2)

PR #199, opened 2016-12-27
Add support for filemagic

rholloway:
Add support for python environments using filemagic instead of python-magic. These two cannot coexist in the same environment or virtual environment, causing issues using Androguard if filemagic is already being used.


rholloway:
This should also fix the closed issue #124 as well.

PR #200, opened 2016-12-27
Add Support For Filemagic

rholloway:
This PR allows for filemagic support in addition to python-magic. This is needed for environments already making use of filemagic, because python-magic and filemagic cannot be used in the same environment or virtual environment.

This PR also addresses the issue raised in #124.


eighthave:
would be nice to test this in travis, but looks simple enough.

PR #202, opened 2016-12-30
remove inline assembly in RDTSC: now CLOCK_MONOTONIC

FauxFaux:
The inline assembly in this function:

 * breaks compiling on non-x86 platforms like ARM
 * probably doesn't work on multi-core systems in general (i.e. most x86 computers now)
 * probably doesn't work on anything with power saving
 * probably doesn't work at all on x86_64, as the `"=A"` might be wrong
 * doesn't seem to be actually used anywhere

Please either remove it, or replace it with `clock_gettime(CLOCK_MONOTONIC, ..`, which is at least, not broken.


adesnos:
Thank you for the commit, but elsim is no more supported

PR #208, opened 2017-02-22
Use python2/3, add documentation, fix small issues, remove old code

reox:
Hi!

I tried to upgrade the whole androguard to be able to work with both python2 and python3.
The main difference is, that I use bytearrays instead of strings, as they are compatible with both versions and in fact the underlying data structure is bytes (e.g. DEX files).
The testcases run, except for one in DAD where I'm not sure if this is caused by lists vs sets, see the FIXME in androguard/decompiler/dad/tests/dataflow_test.py
I think the problem here is, that actually a set should be used instead of lists, as the order of the items does not matter. But with the new hashing randomization in python, the order of dictionary keys is more random now, which causes the UD / DU dicts to be randomized. Therefore the resulting lists will have a different order. I could see, that sometimes the testcases actually work - which is a clear indicator for this behaviour.

In the GUI I needed to introduce a small "hack" as QTreeWidgetItems are somehow not hashable in python3 - which is probably a bug.

I also changed the certificate parsing in apk.py from chilkat to pyasn1 and cryptography. One problem here is, that certificates that do not have a Issuer or Subject correctly formatted by RFC5280 - e.g. country code not two chars - this method will fail. Chilkat seems to be more tolerant to "wrong" certificates.
Maybe we could add a flag to change back to chilkat if necessary? The reason for the change is, that pyasn1 and crytography are widely available (e.g. androguard is now in debian!), where chilkat isn't.

I tested the whole thing by using several thousand APKs, opened them with AnalyzeAPK and called show methods on the resulting apk and dex objects. I tried to minimize the errors as much as possible.
The compatibility to python2 should be maintained, the biggest difference is that bytes should be prefered, but if you are still using python2, I found that everything should behave the same.

I also added the files for sphinx to build the documentation and added some more hints for building in the Readme.

It would be really really great if more people than just me could test the code and give feedback! 

Unfortunately I did not took a look at elsim, so elsim is still python2 only. 
I also found that most of the demos are broken, maybe one should remove this folder and add some new demos?


eighthave:
This sounds great!  Python3 support is the blocker for switching [fdroidserver](https://gitlab.com/fdroid/fdroidserver) to androguard from our own custom crappy `aapt dump` parser.  I'm going to test this now.

I'll update the Debian package with this, once its merged.  It would be good to also have feedback from @appknox @dhilipsiva @subho007 since they are currently maintaining a py3-only port.  Ideally we can merge efforts again with this pull request, since it provides both py2 and py3 support.

subho007:
That's great @reox for the work. Although I am uncertain if @adesnos is going to merge any PR's yet. 

Let me review this PR 👍 

Thanks for the mention @eighthave 😄

reox:
I found out, that the decompilation would not print source code anymore.
I assume this is due to some str/bytes/... conversion problems inside androguard.decompiler.dad.writer.
Somehow the exceptions that should be raised are not propagated to the user or catched in between.
I'm currently investigating this, but I think that the writer needs some special attention.

subho007:
We can create an organization and fork androguard there with a community fork. I am up for it.

adesnos:
I'm not dead ! I can gave access to the repository if you wish

leonid-s-usov:
Though I have 0 minutes 1 second a week currently to review the status of the project I would also vote for forking it. 

Since the times I was in a different company and used and contributed to androguard I felt like the author is not willing to continue as they started - for any valid reason they might have for it. I don’t mean to offend or blame anyone - just putting it as I see it.

adesnos:
I added reox/subho007/eighthave to be able to commit. Could you check please ?

It is an important commit, we need to check that carefully.

adesnos:
@leonid-s-usov yep it was true. I will time to spent more time on it !

adesnos:
s/time/try

subho007:
Thanks @adesnos 👍 I still don't have write access yet

reox:
@adesnos nice to hear! and thank you for adding us! I hope I can contribute to the project in the future the same way as the last weeks.

Yes, it is true: these commits could break things. I hope that everything is still fine, but a large project of me that depends on androguard uses still the version from mercurial and thus I can not check if it still works (I'm currently rewriting it to use the newest version)...

The changes to the certificate parsing in apk.py are certainly a thing which breaks the API.

subho007:
I can commit sat/sun entirely to androguard. 1st thing to go forward is to set-up a proper CI test for existing functions

adesnos:
@subho007 You have not yet accept the invitation

@reox yep. We should add more tests to be sure that we will not break everything

eighthave:
I now have an "Merge pull request" button, so I have write access.
Thanks :).  We should discuss commit/merging procedures. My guess is
that we mostly are interested in maintenance rather than rapid
development.  I like this basic workflow:

* all code through pull requests
* pull requests should be merged by someone else besides the author
* waiting a bit for more reviews is preferrable to rapid mergin

reox:
I found an issue regarding the encoding:
I have an APK containing support libs, and with this file, there are some issues https://github.com/android/platform_frameworks_support/blob/d6d7055c3064a711ef5311b23f57f752a2560a79/compat/java/android/support/v4/util/PatternsCompat.java
There are unicodes in there, that python will not encode (even with `surrogateescape` encoder):
```
>>> "\ud800".encode("utf8", "surrogateescape")
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: 'utf-8' codec can't encode character '\ud800' in position 0: surrogates not allowed
```

It can be easily circumvent by using the `unicode-escape` encoder and then applying the `ascii` decoder.
I think this should be done in the decompiler itself, so all strings inside the source would be then shown as an escaped sequence.
I'm not sure, how this was before? What do you think?

adesnos:
I will merge it and fix issues later :)

adesnos:
@reox could you fix the conflict due to my last commit ?

reox:
Yes, i'll commit that soon

reox:
btw, would you consider to put elsim into another repo and git submodule the right androguard version so people who are interessted in it can still use it?

adesnos:
Yes probably but I want to focus only on Androguard right now and no other dependencies

adesnos:
Thank you for to resolve the conflict I will merge that !

adesnos:
@reox I got the following error with pyhon3.6 now:
androguard/androguard/core/bytecodes/apk.py", line 4, in <module>
    from future import standard_library
ModuleNotFoundError: No module named 'future'

adesnos:
ok we need to add the dependencies in the doc/etc:
future
pyasn1

reox:
yes, i hope i documented them. they also should be in the setup.py.

I think for the travis CI it would be good to add an additional py3 task

PR #217, opened 2017-03-16
setup standard install paths for GUI files

eighthave:
This should install the GUI files when installed using `pip install` so that the files are found.  This should also still work when run straight from git.  Yes, the setting of data_prefix is not pretty, but I have never found a better way...


adesnos:
You did that in the case you installed Androguard (I never did it :))

eighthave:
The idea here is mostly to make it easier for people including androguard in packaging systems like pip/pypi, Debian, Ubuntu, Fedora, Homebrew, etc.  Otherwise, each package will have to have a custom version of this.

The CI fail is from the 3.6 build, which didn't work before this pull request.  2.7 is passing.

reox:
+1

PR #228, opened 2017-04-06
fix some unicode bugs with the decompiler

reox:
Some unicode and byte conversions where overseen because all exceptions where caught. This should fix the problems with not completly decompiled source.

### DIFF
--- a/androguard/core/bytecode.py
+++ b/androguard/core/bytecode.py
@@ -223,8 +223,7 @@ def method2dot(mx, colors={}):
     registers = {}
     if method.get_code():
         for DVMBasicMethodBlock in mx.basic_blocks.gets():
-            for DVMBasicMethodBlockInstruction in DVMBasicMethodBlock.get_instructions(
-            ):
+            for DVMBasicMethodBlockInstruction in DVMBasicMethodBlock.get_instructions():
                 operands = DVMBasicMethodBlockInstruction.get_operands(0)
                 for register in operands:
                     if register[0] == 0:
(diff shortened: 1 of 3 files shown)


frennkie:
I encountered this bug while working on Viper and can confirm that this PR fixes this issue.

Rafiot:
Any update?

frennkie:
@adesnos Can this PR be merged? Or is there something that needs to be done still? It would really helpful if this could get into master.

reox:
@frennkie it would be good to check the whole code for unicode based errors, as there might be some more.
Does anyone has a good tool to check for such errors? 

I think it is possible to merge this and create new errors in the issue tracker.

frennkie:
I think the best way to deal with this in general is to have a broad coverage with unittests (and if possible integration tests) running on both Python2 and Python3. For unicode issues you could add some tests that specifically target such use cases and files.

Running `find ./ -type f -iname "*.py" -exec py3kwarn {} \;` might help to identify and fix some (py3) issues (although IMHO you can not blindly trust the suggestions py3kwarn gives).

reox:
yes, test coverage would be really nice, not only because of the unicode errors...

At the moment I found the errors by running analysis over a few hundred files, which could be potentially automated but takes ages to run...

frennkie:
Well, you already have Travis setup. What @Rafiot and I introduced to Viper is implementing the tests with `pytest` and running them using tox (which make it's easier to run several environments in a convenient and reproducible way - both on Travis and I you like locally).

Issue #239, opened 2017-05-23
travis python2.7 is failing because of ipython dependency


reox:
Looks like ipython 6 dropped python2 support and travis does not like to install something else:
```
Collecting ipython>=5.0.0 (from androguard==2.0)

  Downloading ipython-6.0.0.tar.gz (5.1MB)

    100% |################################| 5.1MB 108kB/s 

    IPython 6.0+ does not support Python 2.6, 2.7, 3.0, 3.1, or 3.2.

    When using Python 2.7, please install IPython 5.x LTS Long Term Support version.

    Beginning with IPython 6.0, Python 3.3 and above is required.

    See IPython `README.rst` file for more information:

        https://github.com/ipython/ipython/blob/master/README.rst

    Python sys.version_info(major=2, minor=7, micro=9, releaselevel='final', serial=0) detected.

    Your pip version is out of date, please install pip >= 9.0.1. pip 6.0.7 detected.

    Complete output from command python setup.py egg_info:

    

    

    IPython 6.0+ does not support Python 2.6, 2.7, 3.0, 3.1, or 3.2.

    

    When using Python 2.7, please install IPython 5.x LTS Long Term Support version.

    

    Beginning with IPython 6.0, Python 3.3 and above is required.

    

    

    

    See IPython `README.rst` file for more information:

    

    

    

        https://github.com/ipython/ipython/blob/master/README.rst

    

    

    

    Python sys.version_info(major=2, minor=7, micro=9, releaselevel='final', serial=0) detected.

    

    Your pip version is out of date, please install pip >= 9.0.1. pip 6.0.7 detected.

    

    

    

    ----------------------------------------

    Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-WRMWjs/ipython

The command "pip install ." failed and exited with 1 during .

```

Who has travis$foo and can fix this?

PR #244, opened 2017-06-10
Fix travis build

Rafiot:
Fix https://github.com/androguard/androguard/issues/239

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,11 @@ import os
 
 from setuptools import setup, find_packages
 
+if sys.version_info >= (3,):
+    install_requires = ['pyasn1', 'cryptography>=1.0', 'future', 'ipython>=5.0.0', 'networkx', 'pygments']
+else:
+    install_requires = ['pyasn1', 'cryptography>=1.0', 'future', 'ipython>=5.0.0,<6', 'networkx', 'pygments'],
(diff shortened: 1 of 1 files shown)


Rafiot:
That's what I did in viper requirements file and it fixed it... Let's see.

reox:
This one is weird... It seems that androguard is installed by pip install . - so the setup.py is executed? Maybe you need to specifiy the version for py2.7 there?

Rafiot:
Yep, I just tried an other fix in setup.py

Rafiot:
(okay, I'll test locally and stop pushing untested thing...)

reox:
I think you need this patch here (last block of code) https://github.com/ipython/ipython/blob/master/README.rst#ipython-requires-python-version-3-or-above

Rafiot:
Right, that's pretty much what I did ;)

Rafiot:
(now fails on python 3 for an other reason)

reox:
yes, but I think you fixed it! The py3.6 error was there before... Somehow there is now some object which can not be serialized... but this is another issue.
Thanks for looking into this!

Rafiot:
You're welcome, I saw this issue while looking for opening an other one ;)

Issue #253, opened 2017-06-17
EncodedField and EncodedMethod have no method '_reload'


cdkorzen:
### Describe what you wanted to do
I wanted to call the method `set_name()` on an EncodedField, and later on an EncodedMethod, to set the in-memory name for a method that had been set to a single-letter, presumably by ProGuard.

- is this error specific to a single file?
Yes, andguard/core/bytecodes/dvm.py

- if possible give a minimal working example where the error happens
Presuming you have an APK loaded as `a`, the DEX loaded as the DalvikVMFormat `d`, and that the APK has a class `ExampleClass` with field `a`

```python
d.C.LExampleClass.F.a.set_name('new_name')
```

### Describe what you expected
The field referred to as `d.C.LExampleClass.F.a` is now referenced as `d.C.LExampleClass.F.new_name` and shown as `new_name` whenever the field is used, such as when calling `str()` or looking up the Xrefs

### Describe what actually happened
```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-36-5f433d2fe809> in <module>()
----> 1 d.C.La.F.a.set_name('new_name')

C:\Android\re_tools\AndroguardMaster\androguard\core\bytecodes\dvm.pyc in set_name(self, value)
   2786     def set_name(self, value):
   2787         self.CM.set_hook_field_name(self, value)
-> 2788         self._reload()
   2789 
   2790     def get_obj(self):

AttributeError: 'EncodedField' object has no attribute '_reload'
```

### System Information

- Androguard Version: master
- Python Version: 2.7 Anaconda
- Operating System: Windows 7 64-bit Home

### Further Log Files and Output

- please report logs, tracebacks and other useful information
- if possible report the files you tried to analyze!

commit 9d519eaae6542 changed the call from `self.reload()` to `self._reload()`  EncodedField and EncodedMethod still have `reload` methods.  This same commit added `_reload()` methods to other classes in the file, such as `FieldIdItem` and `MethodIdItem`.  However, `EncodedField` and `EncodedMethod` did not have their reload functions changed.

reox:
@cdkorzen can you check if the PR resolves the issue for you?
thanks in advance!

PR #255, opened 2017-06-19
Fix renaming methods

reox:
This commit should resolve issue #253 
@cdkorzen can you check if it does? I tried your code and created a test case, which works now fine!


cdkorzen:
Thanks, it works great for me!

PR #256, opened 2017-06-21
Cosmetic Changes and Documentation

reox:
Start on cleaning up the code and write documentation.


reox:
interesting, looks like I hit another version specific python thing with cyclic imports

PR #259, opened 2017-06-22
fix APK checks for unsigned apks, resolves #172

### LINKED ISSUE #172: command "a,d,dx = AnalyzeAPK(apkname,decompiler)" cannot get the dx
Hi,
I used the script androlyze.py in androguard, after I entered the cmd, running the command "a,d,dx = AnalyzeAPK(apkname,decompiler)", i should get the value of a, d, and dx, but the result showed to me is that the d and dx is null. And I cannot use the "VMAnalysis" to get the analyzed dex object.
Any idea?
Thanks!

reox:
I changed the way APK files are detected, as there are also unsigned apks.

This should fix the bug in #172 if an unsigned APK is used.


cdkorzen:
I'll try to check this out tomorrow.  

In the commit there seems to be a lot of changes related to `get_permissions()`, is that intended?

reox:
the functions are duplicated, so i replaced one of them.

cdkorzen:
Just wasn't sure if the changes for permissions were for the unsigned APKs or were meant for a different commit/pull-request.

Regardless, I was able to open an unsigned APK, when I used to get something like the error mentioned in #172

Issue #235, opened 2017-04-28
DAD generates incorrect graph and uses wrong node for a switch's follow property


cdkorzen:
I'm reverse engineering a large function, and DAD is using the wrong node to follow the switch.

The method is essentially a resource loader.  It does a switch on one of its arguments, and then in most or all of the cases tries to open a file.  It reads/processes that file, then breaks.  Before the function exists, it tests if the file variable is not null, and if so closes the file, otherwise just exits.

Expected source should look something like this:
```
{
    switch (p13) {
        case 1:
              \\do stuff
              break;
        case 2:
              \\do other stuff
              break;
        case 3:
              \\do different stuff
              break;
        default:
              \\do default stuff
              break;
    }
    if (file not null) {
        file.close();
    }
    return;
}
```

What the source looks like is that only in the first conditionals and the first switch case is the file variable tested, and closed if necessary.  When elaborating over the other cases, it sees that it already visited the test/close nodes and so ignores them.  I've tried reproducing it in a small example, but I'm unable to do so.

Here's a bit of what it looks like, from a subclassed writer to try and capture the node flow:
```
    void r(int p13, int p14, int p15)
    {
\\Visiting node 2-Switch(r-BB@0x0)
        v0_0 = 0;
        switch (p13) {
            case 0:
\\Visiting node 475-If(r-BB@0x12a)
                if (p14 != 0) {
\\Visiting node 492-If(r-BB@0x35c)
                    if (p15 != 11) {
\\Visiting node 493-SC(500-If(r-BB@0x3be) || 501-If(r-BB@0x3c6))
                        if ((p15 < 13) || (p15 > 18)) {
\\Visiting node 510-If(r-BB@0x5de)
                            if (p15 != 19) {
\\Visiting node 511-If(r-BB@0x5fe)
                                if (p15 != 20) {
\\Visiting node 512-If(r-BB@0x61e)
                                    if (p15 != 21) {
\\Visiting node 513-If(r-BB@0x63e)
                                        if (p15 != 22) {
\\Visiting node 514-If(r-BB@0x65e)
                                            if (p15 != 10) {
\\Visiting node 515-Statement(r-BB@0x67e)
\\Starting analysis of node 515-Statement(r-BB@0x67e)
\\Successors: [522-If(r-BB@0xa)]
                                                v0_0 = this.getResources().getAssets().open(new StringBuilder(String.valueOf(this.gO[p15])).append(".img").toString());
[...]

PR #263, opened 2017-06-23
Fix decompiler simplification

### LINKED ISSUE #262: Question regarding change in decompiler output between v2.0 and latest
When I ran the decompiler after switching from the 2.0 tagged release to the master, the output changed fairly radically.  It's not apparent to me just why that happens.

It seems that DAD was updated to be much more literal.  Compound expressions are broken down and the version attachment to the variables is gone.  It can sometimes affect readability, like how 
in v2.0:
```
                byte[] v4_8 = ((v3_5[v1_19] & 255) | ((v3_5[(v1_19 + 1)] & 255) << 8));
```
now is
```
                    v4 = v3[v1];
                    v4 &= 255;
                    v5 = (v1 + 1);
                    v5 = v3[v5];
                    v5 &= 255;
                    v5 <<= 8;
                    v4 |= v5;
```

Any thoughts?

reox:
It looks like the use of generators was problematic for the simplification of code.

This PR should resolve #262


cdkorzen:
#262 looks fixed for what I was using.  #235 is still there, I'll look into it when I can.

PR #278, opened 2017-07-10
Add skip files parse param

corcoran:
The get_files_types method is the most CPU intensive part of parsing an APK and on some bigger APKs can take minutes depending on your machine.  This param lets you optionally skip it.

### DIFF
--- a/androguard/core/bytecodes/apk.py
+++ b/androguard/core/bytecodes/apk.py
@@ -248,7 +248,6 @@ class APK(object):
 
                         self.valid_apk = True
 
-            self.get_files_types()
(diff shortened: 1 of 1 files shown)


reox:
It is probably not even necessary to have this flag, just remove the `self.get_files_types()` from `__init__` - the file types should then be generated on the first call of `get_file_types`.

What do you think?

corcoran:
Yea that works too, I had just assumed you would want to keep the default behavior.

I can update the PR with that change if it's what's preferred.

corcoran:
@reox Force updated the branch.  How does that look?

reox:
looks good, thanks!

PR #287, opened 2017-08-16
check magic file before using it

iNarcissuss:
The "magic_file" field should be checked like the code snippet above

- Androguard Version: master
- Python Version: 2.7 (Anaconda32)
- OS: Windows 7 64bit


reox:
can you merge with the current master, so travis will run again? thank you!

PR #288, opened 2017-08-20
.encode('utf-8') instead of str()

Airy-Fairy:
Using str() causes UnicodeEncodeError in get_elements(), hence in get_activities(), get_services(), etc.
image
image


reox:
can you merge with the current master, so travis will run again? thank you!

Issue #280, opened 2017-07-31
AnalyzeAPK() for some APKs gives:  struct.error: unpack requires a bytes object of length 4


wdormann:
### Describe what you wanted to do

When parsing https://play.google.com/store/apps/details?id=com.ovkuse.appsbunch with Androguard's androlyze.py, I get:  struct.error: unpack requires a bytes object of length 4
Some other apps give the same error:
org.iShia.iShia.Free
com.app.cp.blog8066
com.app.cp.blog2585
etc...

### Describe what you expected
Parse w/o error

### Describe what actually happened

Androguard version 3.1.0-dev
In [1]: a,d,dx = AnalyzeAPK('com.ovkuse.appsbunch.apk')
WARNING: invalid chunk type
NoneType
---------------------------------------------------------------------------
error                                     Traceback (most recent call last)
/usr/local/lib/python3.5/dist-packages/androguard-3.1.0.dev0-py3.5.egg/EGG-INFO/scripts/androlyze.py in <module>()
----> 1 a,d,dx = AnalyzeAPK('by_sha256/01/54/43/015443accccda1dbca5513ad8196a9c52870c8905b34985e0eb9c27a0220c6c2/com.ovkuse.appsbunch.apk')

/usr/local/lib/python3.5/dist-packages/androguard-3.1.0.dev0-py3.5.egg/androguard/misc.py in AnalyzeAPK(filename, session)
     42         data = fd.read()
     43
---> 44     session.add(filename, data)
     45     return session.get_objects_apk(filename)
     46

/usr/local/lib/python3.5/dist-packages/androguard-3.1.0.dev0-py3.5.egg/androguard/session.py in add(self, filename, raw_data, dx)
    152             digest = hashlib.sha256(raw_data).hexdigest()
    153             if ret == "APK":
--> 154                 apk_digest, apk = self.addAPK(filename, raw_data)
    155                 dex_files = list(apk.get_all_dex())
    156

/usr/local/lib/python3.5/dist-packages/androguard-3.1.0.dev0-py3.5.egg/androguard/session.py in addAPK(self, filename, data)
     72         digest = hashlib.sha256(data).hexdigest()
     73         androconf.debug("add APK:%s" % digest)
---> 74         apk = APK(data, True)
     75         self.analyzed_apk[digest] = [apk]
     76         self.analyzed_files[filename].append(digest)

[...]

Issue #283, opened 2017-08-15
Some AXML files start with another header


reox:
It seems some AXML files (usually the AndroidManifest.xml), do not start with the usual AXML Header of `0x00080003` but with `0x00080503`. 
One solution is to allow this other header as well:
```
diff --git a/androguard/core/bytecodes/apk.py b/androguard/core/bytecodes/apk.py
index 1d4d780..458a76a 100644
--- a/androguard/core/bytecodes/apk.py
+++ b/androguard/core/bytecodes/apk.py
@@ -1222,7 +1222,7 @@ class AXMLParser(object):

         axml_file = unpack('<L', self.buff.read(4))[0]

-        if axml_file == CHUNK_AXML_FILE:
+        if axml_file == CHUNK_AXML_FILE or axml_file == 0x00080503:
             self.buff.read(4)

             header = ARSCHeader(self.buff)
```

Who has more knowledge of the AXML format and might answer the question with the different headers?
If you search for 0x00080503 on github you find nothing useful... It seems like nobody uses this constant? Where does it comes from?

reox:
Some update on that: I scannen 50000 APKs and read the first four bytes of AndroidManifest.xml.
I found 319 files that had another header, namely:
```
      1  0x03570800
      1  0x88330800
      3  0x01000800
      3  0x03010800
      3  0x33330800
      4  0x00010800
      4  0x00020800
      5  0x00030800
      5  0x03020800
     11  0x02000800
    279  0x00000800
```

I did not tried to install them on a device (maybe I should check if those files are valid) but it looks like there are a lot of packers/obfuscators that change the header.
By the looks of it, the only thing that is always the same is the 0800.

I'll change the check, so such files are at least tried to read.

@androguard/external-contributors anyone has more information about the AXML file or this specific case? The documentation is really really sparse...
There are also other tickets regarding the AXML parser, does anyone has experience with this format?

reox:
So I adopted the code, but it looks like a lot of those files are not parsable. I found 75 that would work and 243 produced other errors...

If someone wants to tackle this problem, I can provide the set of files I found.

Issue #281, opened 2017-07-31
AnalyzeAPK() for some APKs give a variety of xml.parsers.expat.ExpatError: errors


wdormann:
### Describe what you wanted to do

When parsing https://play.google.com/store/apps/details?id=com.shopgate.android.app13182 with Androguard's androlyze.py, I get: xml.parsers.expat.ExpatError: unbound prefix: line 108, column 0

Other apps have similar errors in expat, but it's not clear if the underlying cause is the same.  The whole set of unique expat errors I've seen is:
$ grep ExpatError invalid.err | sort | uniq
xml.parsers.expat.ExpatError: duplicate attribute: line 101, column 0
xml.parsers.expat.ExpatError: duplicate attribute: line 110, column 0
xml.parsers.expat.ExpatError: duplicate attribute: line 121, column 0
xml.parsers.expat.ExpatError: duplicate attribute: line 164, column 0
xml.parsers.expat.ExpatError: duplicate attribute: line 81, column 0
xml.parsers.expat.ExpatError: duplicate attribute: line 98, column 0
xml.parsers.expat.ExpatError: must not undeclare prefix: line 2, column 0
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 154, column 0
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 205, column 6
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 338, column 6
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 349, column 6
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 397, column 6
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 509, column 6
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 569, column 6
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 57, column 0
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 5, column 24
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 602, column 6
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 611, column 6
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 614, column 6
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 629, column 6
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 642, column 6
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 670, column 6
[...]

PR #289, opened 2017-08-29
Fix axml missing ns / extra style offset

reox:
This PR should fix the issues #280 and #281.

There are also extra unit tests for those cases.


reox:
this one should also fix #283

PR #293, opened 2017-08-31
test if tests on osx work as well

reox:
may test more stuff in this PR


reox:
looks like osx builds are broken on travis... leave it as it is

PR #340, opened 2017-11-07
use generator as list again to access last element

reox:
maybe rework the function later...


reox:
-.- pycrytography seems to be broken for py3.3 *grrrr*

PR #345, opened 2017-11-20
Run tests with PyPy 5.8

thedrow:
Since this is a pure python library I see no reason why PyPy shouldn't be supported.
Let's see if this is a no brainer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ python:
   - "3.5"
   - "3.6"
   - "3.7-dev"
-  - "pypy"
+  - "pypy2-5.9.0"
 # command to install dependencies
 install:
   - "pip install ."
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ python:
   - "3.5"
   - "3.6"
   - "3.7-dev"
+  - "pypy"
 # command to install dependencies
 install:
   - "pip install ."


thedrow:
The failure seems legit since those classes don't have a `get_ref_off()` method.
I'm not sure how they appear on CPython.
Can anyone share their insights?

thedrow:
I just run the tests with PyPy 5.9.0 and all tests pass.

thedrow:
Seems like PyPy 5.9.0 is not available in Travis :(

reox:
I tried pypy some time ago and had the same behaviour... 
Probably we should fix this bug. But I have no idea what pypy does that python don't...

reox:
this looks really weird... there is this check:
https://github.com/androguard/androguard/blob/67b992ce0feeeb01bc69a99257916487689c3bcf/androguard/core/analysis/analysis.py#L137

And the opcodes 0x26, 0x2b, 0x2c have all the `t` format. I'm also not sure why not the other `t` types are in this function, but it might have a reason...
So I really have no idea how it is even possible that an other instruction get's into this function?!

reox:
Is it possible that pypy < 5.9.0 has issues with interval compares?!

reox:
see also #347 there are more pypy versions.

@thedrow have you any idea what might be wrong? The interesting thing is, that those errors should not happen. But I never used pypy - have you any experience?

reox:
this is really weird... I tried to implement the method in the base class, but then the exception is raised. Looks like this method is actually called! (See https://travis-ci.org/androguard/androguard/jobs/306241214#L553)

thedrow:
PyPy 5.8.0 had a bug regarding to ordering which was fixed in 5.8.1.
I recommend supporting the latest version of PyPy.

reox:
I think this is partly not the fault of pypy... we had such bugs already where the ordering of dict keys where assumed but in fact they are not ordered.
for this other bug I'm not sure what to do. interestingly it works with some versions.

reox:
I'll close thise one here. There are tickets describing the problem #349 and #350

PR #357, opened 2017-11-29
remove old docs and add docs building to travis

reox:


reox:
lame... this PR shows that PyQT5 is not available for python < 3.5.

reox:
wow... this is harder than I though... On python 3.3 sphinxcontrib.programoutput seems to be broken

Issue #355, opened 2017-11-27
apk.new_zip only adds new_files to a zip if deleted_files is not None


diarmaid-mcmanus:
Hello, I'm unsure if this is a bug or intended feature. Please tag this with question, I don't seem to have access to tags.

When calling `apk.new_zip('example.apk', new_files={'Anything':'content'})` file 'Anything' does not get written. 
file 'Anything' only gets written if deleted_files='no-thing' i.e. is any string other than the file name in new_files. In this example, deleted_files must be something other than 'Anything'.

https://github.com/androguard/androguard/blob/9f88ec774ec9b089f060e075785fe091465d8e22/androguard/core/bytecodes/apk.py#L870

I would expect one of the following two things to happen:

1. new_files and deleted_files to work independently, or
2. new_files to work if the new_files file is also in the deleted_files string.

My preference would be for 1 - if deleted_files=None, then new_files are still written into the zip. Existing files are overwritten.

I could understand 2 - if a file is to be deleted, then it's overwritten with the file in new_files.

I don't understand why it works as in the code at the moment - new_files are written to the zip, if deleted_files is not None, and if the file name in deleted_files does not match new_files.

As on my other issue happy to provide a patch for this, once I know what the expected behaviour is.

reox:
I never used the function :D But it looks like either the documentation is wrong or the implementation.
I think it is the implementation :) I think you should be able to create new files with a given dict of files and also remove files from the APK if necessary.

I would be glad if you can provide a patch!

diarmaid-mcmanus:
Current plan to replace the code at https://github.com/androguard/androguard/blob/9f88ec774ec9b089f060e075785fe091465d8e22/androguard/core/bytecodes/apk.py#L867

```python
for item in self.zip.infolist():
    if deleted_files is not None:
        if re.match(deleted_files, item.filename) is None:
            buffer = self.zip.read(item.filename)
            zout.writestr(item, buffer)
    if new_files is False:
        if item.filename in new_files:
            zout.writestr(item, new_files[item.filename])
zout.close()
```

I haven't tested that the new_files conditional block will overwrite existing files, I'll do this tomorrow.

Issue #354, opened 2017-11-26
apk.show() isn't returning intent filters


diarmaid-mcmanus:
### Describe what you wanted to do

By calling androguard/androguard/core/bytecodes/apk.APK('file.apk').show(), no intent-filters are returned in the printout.

- is this error specific to a single file?
No
- if possible give a minimal working example where the error happens
For use in androlyze.py -s
`from androguard.core.bytecodes import apk
a = apk.APK('example.apk')
a.show()`

### Describe what you expected
Intent filters to be printed after the name of activities, etc

### Describe what actually happened
They're not printed.

### System Information

- Androguard Version: master
- Python Version: 3.5.2
- Operating System Ubuntu 16.04.3

### Resolution
The following line submits bytes to get_intent_filters. These bytes need to be decoded with .decode
https://github.com/androguard/androguard/blob/9f88ec774ec9b089f060e075785fe091465d8e22/androguard/core/bytecodes/apk.py#L986

I'll submit PR and fix for this and some other issues this or next week.

diarmaid-mcmanus:
Should .decode for this live within show() or should there be a sanity check in get_intent_filters? This affects not only activities (as my example) but also services, receivers

reox:
maybe the xml parsing has some issues - i changed it to lxml lately.

It would be very nice if you could provide a fix!

reox:
Looks like this is a python3 issue: I think decode should be already in the a.get_activities() function.

If you have some nice AndroidManifest.xml files with a lot of different intent filters, they would also be well suited for a unittest.

diarmaid-mcmanus:
I can create one. I have an AndroidManifest.xml with such but it's a random manifest from the play store. Is this OK to use?

I plan to fix this and provide a patch for something else I think is a bug (question incoming) this week, hopefully early this week but Sunday at any rate.

diarmaid-mcmanus:
Hey, the following line turns an str into a byte array. Do you know/remember why it was implemented like this? 

https://github.com/androguard/androguard/blob/9f88ec774ec9b089f060e075785fe091465d8e22/androguard/core/bytecodes/apk.py#L538

By removing .encode() on this line it works as I'd expect, and all test_*.py files pass. I'm not sure if that's appropriate as a fix though - will it break something else?

PR #358, opened 2017-11-29
fixing #354 with unit tests and #355

diarmaid-mcmanus:
#354 resolved with changes on line 538 of apk.py. Tests added in test_apk.py to test that a string type is returned for get_activities(), which currently returns a byte class.

#355 resolved with moving the new_files check outside of the deleted_files check. No tests for this, and I haven't added any in this PR.


reox:
very nice, thank you!

PR #366, opened 2017-12-04
Added switch for raw in AnalyzeAPK

subho007:
Filename or the the filename of the android application or a buffer which represents the application

### DIFF
--- a/androguard/misc.py
+++ b/androguard/misc.py
@@ -25,7 +25,7 @@ def get_default_session():
     return androconf.CONF["SESSION"]
 
 
-def AnalyzeAPK(filename, session=None):
+def AnalyzeAPK(filename, session=None, raw=False):
     """
         Analyze an android application and setup all stuff for a more quickly analysis !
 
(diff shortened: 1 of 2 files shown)


subho007:
@reox good to merge ?

androguard/misc.py
```diff
@@ -40,8 +40,11 @@ def AnalyzeAPK(filename, session=None):
...
 
-    with open(filename, "rb") as fd:
-        data = fd.read()
+    if raw:
+        data = filename
+    else:
+        with open(filename, "rb") as fd:
+            data = fd.read()
 
     session.add(filename, data)
```
reox: if you supply raw data, the filename will be set with the raw data... this is probably not good...
maybe calcuate a hash over the raw data first and use it as filename? dunno..

otherwise it looks good!

reox:
maybe change that one line to have a shorter human readble string in filename? everything else looks good!

subho007:
@reox on it 👍

PR #370, opened 2017-12-11
Always return an Array of DalvikVM objects to support multi-dex app

### LINKED ISSUE #369: Return array for both single or multi dex for `AnalyzeAPK`
### Describe what you expected

`_a, _vm, _vmx = AnalyzeAPK(file_content, raw=True)`

I expect _vm to be an array for both single or multi dex. For single dex the array for be having a single object.

### Describe what actually happened
`_vm` is an object if single dex and array for multi dex

subho007:
This will return an Array of DalvikVM objects for both multi and single dex application in `AnalyzeAPK` method.

This PR has the following changes:

- AnalyzeAPK will return `APK` object, list of `DalvikVM` objects and `Analysis` object
- Previously for multi-dex application an array of `Analysis` objects were returned which is wrong since the Analysis will have reference of all the `dex` files via the `add()` method
- Fix the test cases to accommodate with the new return types of `AnalyzeAPK`
- Add test-cases and sample apk for multi-dex app from https://github.com/subho007/abcore (forked version of https://github.com/greenaddress/abcore)

Fixes #369


subho007:
@adesnos @reox let me know if this looks good

reox:
probably we can test now decompilation of all dex files in an APK?
I also noted that we should check if the xrefs do work when using multiple dex files... It would be very nice to have a very small multidex apk, with only a few classes per dex, which are referencing each other.

But the coomit looks good to me!

subho007:
@reox I'm currently investigating on it. In future I'll make a sample smaller multidex application and write some more testcases on it. Have created a new issue #371

PR #375, opened 2017-12-14
Security fix for XML external entities

bernhl:
See this blog post for reference: https://mikeknoop.com/lxml-xxe-exploit/


reox:
thanks a lot!

PR #376, opened 2017-12-14
Parse TYPE_STRING_DATA_ITEM first

bernhl:
When parsing dex files, there are plenty of warnings:
<pre>unknown string item @ 0x%x(%d)</pre>
This is caused during creation of MapItem (which are in turn created in MapList). If MapItems are parsed before the MapItem of type TYPE_STRING_DATA_ITEM is processed, the calls to get_raw_string fail with the above log message.
In the PR I delayed the parsing of MapItem so TYPE_STRING_DATA_ITEM can be processed first.

### DIFF
--- a/androguard/core/bytecodes/dvm.py
+++ b/androguard/core/bytecodes/dvm.py
@@ -6984,6 +6984,7 @@ class CodeItem(object):
 class MapItem(object):
     def __init__(self, buff, cm):
         self.CM = cm
+        self.buff = buff
 
         self.off = buff.get_idx()
 
@@ -6993,10 +6994,6 @@ class MapItem(object):
         self.offset = unpack("=I", buff.read(4))[0]
 
         self.item = None
-
-        buff.set_idx(self.offset)
-
-        self.next(buff, cm)
 
     def get_off(self):
         return self.off
@@ -7010,9 +7007,12 @@ class MapItem(object):
     def get_size(self):
         return self.size
 
-    def next(self, buff, cm):
+    def parse(self):
         log.debug("Parsing section %s" % TYPE_MAP_ITEM[self.type])
         started_at = time.time()
+        buff = self.buff
+        buff.set_idx(self.offset)
+        cm = self.CM
 
         if TYPE_MAP_ITEM[self.type] == "TYPE_STRING_ID_ITEM":
             self.item = [StringIdItem(buff, cm) for i in range(0, self.size)]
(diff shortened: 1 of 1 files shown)


reox:
Nice! If I remember correctly, there is a reload called on all map items afterwards, we could skip this then, right?

bernhl:
I'll investigate

bernhl:
I dug a bit into the code and found a couple of places which still require a reload. In particular, 

- ProtoIdItem
- FieldIdItem
- MethodIdItem
- EncodedField
- EncodedMethod
- ClassDefItem

have logic such as resolving members (e.g. get_type, get_prototype).
Only TypeIdItem only depends on a string and could be skipped in the future.
I'm wondering why reloading is necessary for DCode, DalvikCode and CodeItem.
@Lircyn: What's the reason for using DebugInfoItemEmpty instead of DebugInfoItem?

reox:
thanks for looking into that! 

Can the dependecies modelled as a directed acyclic graph? I think so, right? Then there is a sequence in which order the map items can be read to avoid any reloading.

For the basic data items it should be:
1. Strings
2. type
3. proto, field and class
4. method

and then there is the rest... but I also need to investigate how it is currently implemented

reox:
Isnt the value of the map_item actually the order in which the items can be parsed? It looks like this order is actually the order we want: https://source.android.com/devices/tech/dalvik/dex-format#type-codes

bernhl:
TYPE_PROTO_ID_ITEM depends on TYPE_TYPE_LIST, so the order of the type code table doesn't give a load order. Also, the description for map_list reads:

> A given type must appear at most once in a map, but there is no restriction on what order types may appear in ...

Still, I built a DAG here: https://github.com/bernhl/androguard/commit/a8ae2ae12b8dc8d2935e35c92f5e04fb82e2e583 and removed the reloading code here: https://github.com/bernhl/androguard/commit/086155bb7811c360431e61787a78d0fb1a1e4143
This is not yet ready to merge as I have to check how that interferes with the reload calls in set_hook_class_name, set_hook_method_name, set_hook_field_name and EncodedField.set_name

reox:
nice work!

> TYPE_PROTO_ID_ITEM depends on TYPE_TYPE_LIST, so the order of the type code table doesn't give a load order.

No, I meant the integer value of those items: type_list is 0x1001 and proto_id is 0x0003
you could check if your DAG gives the same result as the integer ordering of the types.

PR #380, opened 2017-12-17
Change string processing and unify output for py2/3 of get_raw() methods

reox:
All the instructions probably should also return bytearrays...

At least, this program outputs now the same for py2 and 3:
```
from __future__ import print_function
from androguard.misc import AnalyzeAPK
from binascii import hexlify

a,d,dx = AnalyzeAPK("examples/tests/hello-world.apk")

for dvm in d:
    for m in dvm.get_methods():
        code = m.get_code().get_raw() if m.get_code() is not None else bytearray()
        print(m.get_class_name(), m.get_name(),
                hexlify(code).decode("ASCII"))
```

have to use hexlify, as bytearrays are interpreted as str in py2 and bytes in py3.


reox:
this is interesting: I wrote a small dex parser to extract the bytecodes. Then I found out, that there are several classes, where the output of androguard appends something. And there are at least two where a large portion of the bytecode is missing.
This could be harder to investigate what is going on. I skipped the checks for now, but we should definetly see where those differences come from.

We also should do more tests that the dex files are correctly read.

Issue #221, opened 2017-03-17
Taint analysis


adesnos:
Taint analysis could be really useful in Androguard and could reduce significantly an analysis.

Let's do:

- [ ] quick review of the algorithm we want to add
- [ ] first version of a generic support in the analysis to improve/change the algorithm
- [ ] add the corresponding code in the GUI

subho007:
This is awesome. We did had Taint Analysis in Androguard 2.0 why did we remove it in the first place @adesnos ?

reox:
I think the xrefs replaced a lot of it? I rewrote parts of my tools that used tainted to use xrefs which works pretty good (but it is a little bit harder to use...)

hallucino:
@subho007 It was not really good and took a lot of memory.

adesnos:
@subho007 I removed it because it is inefficient, took a lot of memory and we need a clear design

the-house:
@adesnos am happy to contribute to getting taint analysis working again, let me know how I might assist.

subho007:
In the meantime I have ported the old Taint Analysis - https://github.com/appknox/androguard/tree/add-tainted-analysis

commit - https://github.com/appknox/androguard/commit/7af65537ba7051feb67530ab47041985097d5ce6

I think it is better to keep it until a newer implementation is done. Lot of tool depends on the analysis, and it is very difficult to refactor all the tools which uses androguard.

@adesnos @reox thoughts ?

PS: My fork is a re-fork of the current androguard implementation

reox:
Wow, very impressive! I would merge it!

reox:
But is there anything in tainted, that can not be done with the xrefs? I rewrote some parts that used tainted before and it wasn't too hard... 
What are the functions we need?

I think we can do some quick changes to support at least the following things:
* find all calls to / from X
* find all occurences of a string
* find all usages of a package
* ???

subho007:
@reox that is exactly why I didn't send a PR. We can do most of the tainted analysis with XREFs. 

One important thing which we require was `find all instances of Ljavax/crypto/` which is not possible to find from XREF alone, `Ljavax/crypto/` is a system package, it is found in the descriptor of the current method.

reox:
I started to refactor a little bit the Analysis module. If you have nice ideas for the functions we should implement, please share :)

reox:
@subho007 regarding the javax.crypto: if such a module is used, it should end up in the external methods, right? Do you have an APK to test this on? I think this should work already...

[...]

PR #384, opened 2017-12-18
[WIP] Tainted

reox:
starting to do some reformatting on the Analysis module.

This is heavily work in progress! So do not merge...

See #221 for more information


subho007:
I'm starting to fall in love with androguard once again @reox !

reox:
Sooo, I added some find methods, to query the analysis object for certain objects like classes, methods, strings and fields.

Issue #392, opened 2017-12-28
Errors raised during valid APK decoding


U039b:
We use **Androguard** in [εxodus](https://github.com/Exodus-Privacy/exodus) and [εxodus standalone](https://github.com/Exodus-Privacy/exodus-standalone) and it appears that **Androguard** is unable to decode some APKs. 

### What I wanted to do
In εxodus, we need to list Java classes contained into a given APK. To do so, we load, decode and call `get_classes_names()` method.
- This issue can be reproduced by decoding [this APK](https://framadrop.org/r/MWIlNfShVH#7PqXXozRgvcO0W3gslUt6ERxSk2RDMUdytxr8oq3wrU=)

### Error log
```
Traceback (most recent call last):
  File "manage.py", line 22, in <module>
    execute_from_command_line(sys.argv)
  File "/home/exodus/exodus/venv/lib/python3.5/site-packages/django/core/management/__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "/home/exodus/exodus/venv/lib/python3.5/site-packages/django/core/management/__init__.py", line 356, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/exodus/exodus/venv/lib/python3.5/site-packages/django/core/management/base.py", line 283, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/home/exodus/exodus/venv/lib/python3.5/site-packages/django/core/management/base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "/home/exodus/exodus/exodus/reports/management/commands/refreshstaticanalysis.py", line 76, in handle
    static_analysis.save_embedded_classes_in_file(fp.name)
  File "/home/exodus/exodus/exodus/exodus/core/static_analysis.py", line 103, in save_embedded_classes_in_file
    f.write('\n'.join(self.get_embedded_classes()))
  File "/home/exodus/exodus/exodus/exodus/core/static_analysis.py", line 94, in get_embedded_classes
    self.decode_apk()
  File "/home/exodus/exodus/exodus/exodus/core/static_analysis.py", line 51, in decode_apk
    self.decoded = DalvikVMFormat(self.apk)
  File "/home/exodus/exodus/venv/lib/python3.5/site-packages/androguard/core/bytecodes/dvm.py", line 7567, in __init__
    self._load(buff)
[...]

PR #393, opened 2017-12-28
[WIP] on unicode decode error, escape surrogates

reox:
Better: what would be the string?

See #392


reox:
resolve #392 by escaping strings that are problemtic.
Adding an additional get_strings_unicode method to DalvikVMFormat to get the "exact" string, which is UTF-16 coded and surrogates are given as two 16 bit codes.

PR #415, opened 2018-02-12
Get element by attribute name from manifest

ksg97031:
Bug fix

Bug case : `<activity background="\#ffffff" label="x" name="com.x.x.x.NewActivity"/>`

### DIFF
--- a/androguard/core/bytecodes/apk.py
+++ b/androguard/core/bytecodes/apk.py
@@ -528,9 +528,14 @@ class APK(object):
         l = []
         for i in self.xml:
             for item in self.xml[i].findall('.//' + tag_name):
-                value = item.get(NS_ANDROID + attribute)
-                value = self.format_value(value)
+                keys = item.keys()
+                default_key = NS_ANDROID + attribute
+                if default_key in keys:
+                    value = item.get(default_key)
+                elif attribute in keys:
+                    value = item.get(attribute)
 
+                value = self.format_value(value)
                 l.append(value)
         return l
 


reox:
weird, we now have timeout issues with py3.3... I'll close and reopen the PR so travis will run again.

@ksg97031 thanks for the PR! can you write a testcase for this problem, if possible?

ksg97031:
"com.gamevil.psrforkakao" AndroidManifest.xml contains 'name' attribute without 'android' namespace.('**android:name**' -> '**name**')
I think this is a special case cause by android protector so you don't need for care about this issue.

reox:
Okay this is interessting. I downloaded this file: https://koodous.com/apks/853c34b8f18af4eb8de9fc118c0ee31ffad5cd145c36801035a43498a6afd9a1
It looks like the Namespace is set in there, at least the attribute "name" on tag "application" has the namespace uri 30. Just in the beginning, the namespace is started (uri 30, prefix 103 = "android"), so it should generate the prefix correctly.
I checked this APK with the latest master, and the prefix is shown there. Can you hint me to the sample where is this is happening?


btw, it looks like, there is also a bug in the 010 Editor template: https://github.com/strazzere/010Editor-stuff/blob/master/Templates/AXMLTemplate.bt#L239 @strazzere the namespace prefix is not always stored the string item uri-1. The template shows manifest:name for this attribute, but the namespace prefix is actually "android:"

reox:
Ah sorry, I just realized it is not about the application but some activity, like this:
```
<activity name="com.bangcle.everisk.stub.NewActivity" label="EVERISK" background="#ffffff">
</activity>
<activity name="com.bangcle.everisk.stub.InputActivity" label="EVERISK" background="#ffffff">
</activity>
```

I'll investigate that...

reox:
But the raw manifest looks actually correct... the namespace uri is set to 0xFFFFFFFF, which indicates no namespace. Therefore I would not print it...

ksg97031:
Oh sorry.. i test on old version, It's patched now. 👍

PR #433, opened 2018-02-22
Make ASCII detection compatible with Python3

ClaudiuGeorgiu:
In `Python3` you can't call `decode()` on a `str` object, so this workaround should work with both `Python2` and `Python3`.


reox:
maybe this function should be changed to a more robust way... something like this:
any(map(lambda x: ord(x) > 127, string))
but for now this seems fine. Thanks!

PR #436, opened 2018-02-23
working on session

reox:
really returning all classes


subho007:
Any help you need in this ?

reox:
yeah, maybe we should close this one and work on this new issue #437 

I noticed that the tests would fail after the "fix" so maybe it wasn't a fix at all.

PR #454, opened 2018-04-04
Fixed: get_app_icon().

feicong:
get_app_icon() maybe returns xml file,  not a png file.

### DIFF
--- a/androguard/core/bytecodes/apk.py
+++ b/androguard/core/bytecodes/apk.py
@@ -347,7 +347,7 @@ class APK(object):
             try:
                 for config, file_name in candidates:
                     dpi = config.get_density()
-                    if current_dpi < dpi <= max_dpi:
+                    if current_dpi < dpi <= max_dpi and not file_name.endswith("xml"):
                         app_icon = file_name
                         current_dpi = dpi
             except Exception as e:


reox:
thanks for the PR! Do you have some samples that show this behaviour? maybe there is something else wrong.

feicong:
yes, it's an adaptive-icon. the xml file contents like this:

```
<?xml version="1.0" encoding="utf-8"?>
<adaptive-icon
  xmlns:android="http://schemas.android.com/apk/res/android">
    <background>
        <bitmap android:src="@mipmap/adaptiveproduct_google_plus_background_color_108" />
    </background>
    <foreground>
        <bitmap android:src="@mipmap/adaptiveproduct_google_plus_foreground_color_108" />
    </foreground>
</adaptive-icon>
```

when analyzing apk, I usually only want to get the PNG file path with the maximum resolution. If it is adaptive-icon, I need to generate the icon dynamically, which is more difficult to deal with.

here is the test file:

[update.apk.zip](https://github.com/androguard/androguard/files/1886946/update.apk.zip)

subho007:
If the type is XML, one should read the XML and re-call the function to find the icon path.

reox:
that might be very complicated. The background and foreground can be given as some kind of vector file (again an XML).
I think the best solution would be to check in the get_app_icon if a XML is returned, then change the max_dpi to 0xFFFE - 1 (ANY_DPI -1), so you _should_ get an actual icon. The only problem could be apps that are build only for v26. Then there might be no icon file.

feicong:
maybe the best solution is set an argument to the get_app_icon function.

subho007:
what do you think about this @reox ?

reox:
My thoughts on this: Levae the get_app_icon function as it is, as it does the correct thing...
If you really want a png file, then just check it on your own and call get_app_icon with a lower DPI using dpi = 0xFFFE - 1.

There are probably at some day APKs which have only an XML file and no PNG at all. So patching the function to ignore xml at all is probably not so good.

PR #486, opened 2018-04-30
  switch pyasn1 to asn1crypto

shuxin:
so many bugs in pyasn1. asn1crypto is more easy in use, and no need cryptography any more.

### DIFF
--- a/androguard/core/bytecodes/apk.py
+++ b/androguard/core/bytecodes/apk.py
@@ -21,11 +21,7 @@ import lxml.sax
 from xml.dom.pulldom import SAX2DOM
 
 # Used for reading Certificates
-from pyasn1.codec.der.decoder import decode
-from pyasn1.codec.der.encoder import encode
-from cryptography import x509
-from cryptography.hazmat.backends import default_backend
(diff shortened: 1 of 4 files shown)


reox:
This looks really good! Is asn1crypto a drop in for cryptography, ie. do they share the same API? Then it would be totally awesome.

shuxin:
asn1crypto is just a good parser without openssl.
verifing the signature still need openssl/cryptography/M2Crypto/...
(I have wrote a pure python version https://github.com/shuxin/apk-signature-verify, but sitll not support python3 )

reox:
Okay I tested this now a little bit and it looks good! The only thing I can not find is valid_before and valid_after fields, see https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Certificate.not_valid_before
Do you have any idea how to get this information?

shuxin:
type(x509.Certificate.load(c.dump())['tbs_certificate']['validity']['not_after'].native) == datetime

reox:
ok thanks! this is the only thing which is not so straight forward :D

PR #511, opened 2018-05-12
verify sign with python 

shuxin:
apkverify uploaded to https://pypi.org/project/apkverify/


reox:
@shuxin can you fix the bugs? 
There is also this one:
```
  File "/home/travis/build/androguard/androguard/tests/test_session.py", line 47, in testSessionLoad
    session.Save(s, "test_session")
  File "/home/travis/build/androguard/androguard/androguard/session.py", line 27, in Save
    pickle.dump(session, fd)
TypeError: can't pickle _thread.RLock objects
```
I already encounted this one, which comes from pickling zip files. 
Does `self.apksig` in APK contain such things? In that case you need to remove this object before pickling.

PR #525, opened 2018-06-04
fix bug for xref from multi dex

thecrackofdawn:
If an apk contains multiple dex for instance `classes1.dex` and `classes2.dex`. If method `A`  in `classes1.dex` invokes method `B` which is in `classes2.dex`, method `B` will only be searched in `classes1.dex`. So we will get nothing because method `B` is in `classes2.dex`. Instead we should search all classesN.dex for method `B`.

### DIFF
--- a/androguard/core/analysis/analysis.py
+++ b/androguard/core/analysis/analysis.py
@@ -1032,7 +1032,10 @@ class Analysis:
                     if method_info:
                         class_info = method_info[0]
 
-                        method_item = instruction.cm.vm.get_method_descriptor(method_info[0], method_info[1], ''.join(method_info[2]))
+                        for vm in self.vms:
+                            method_item = vm.get_method_descriptor(method_info[0], method_info[1], ''.join(method_info[2]))
+                            if method_item:
+                                break
 
                         if not method_item:
                             if method_info[0] not in self.classes:


reox:
yes, you are right! thanks for the fix!

I also added a short multidex example to test this.
Before:
```
In [23]: list(dx.get_classes())[1].get_xref_to()
Out[23]: 
defaultdict(set,
            {<analysis.ClassAnalysis Ljava/lang/Object; EXTERNAL>: {(1,
               <analysis.ExternalMethod Ljava/lang/Object;-><init>()V>,
               0)},
             <analysis.ClassAnalysis Lcom/foobar/foo/Foobar;>: {(0,
               <androguard.core.bytecodes.dvm.EncodedMethod at 0x7ff8709c2fd0>,
               0),
              (1,
               <analysis.ExternalMethod Lcom/foobar/foo/Foobar;-><init>()V>,
               4),
              (1,
               <analysis.ExternalMethod Lcom/foobar/foo/Foobar;->somemethod(Ljava/lang/String;)V>,
               14)}})
```

After:
```
In [5]: list(dx.get_classes())[1].get_xref_to()
Out[5]: 
defaultdict(set,
            {<analysis.ClassAnalysis Ljava/lang/Object; EXTERNAL>: {(1,
               <analysis.ExternalMethod Ljava/lang/Object;-><init>()V>,
               0)},
             <analysis.ClassAnalysis Lcom/foobar/foo/Foobar;>: {(0,
               <androguard.core.bytecodes.dvm.EncodedMethod at 0x7f4536533748>,
               0),
              (1,
               <androguard.core.bytecodes.dvm.EncodedMethod at 0x7f4536515ef0>,
               4),
              (1,
               <androguard.core.bytecodes.dvm.EncodedMethod at 0x7f4536515f28>,
               14)}})
```

PR #528, opened 2018-06-06
Fix `NoneType` obj has no attribute `findall`.

enkelli:
It is raised from `APK.get_elements()` and `APK.get_main_activity()` when parsing of `AndroidManifest.xml` failed in `APK._apk_analysis()` - `self.xml["AndroidManifest.xml"] is set to `None`.

Same fix is already implemented in `APK.get_element()`.


reox:
thanks for the PR!

PR #546, opened 2018-08-03
Compatibility for Python3 while analyzing ODEX

nbShawnLu:


reox:
does the dex detection not work using strings? I thought this was tested with python 2.7...
there is probaby a different bug than using bytes vs strings.

nbShawnLu:
If not use bytes, it fails to enter the block start with line 8474 in dvm.py: _if self.magic in (ODEX_FILE_MAGIC_35, ODEX_FILE_MAGIC_36, ODEX_FILE_MAGIC_37):_ with Python 3.5.4. It only happens while analyzing ODEX(DEY).

Issue #561, opened 2018-10-02
User experience


MartinThoma:
You use a lot of different command line scripts. As a user, I wish you would use one. I recommend using `click` for that.

reox:
the thing is, that the scripts are usually written for a specific job and used in a chain of other tools. If there would be only a single script, this would probably degrade into something like `androguard <subcommand><options> ...`

Do you think that this would improve the usability? I mean it does in certain cases, for example `virsh` also has subcommands. 

you are very welcome to send a pull request ;)

reox:
I took a look at click and it seems to be quite advanced over the current argument parsers.
What it allows would be to have definitions for commonly used functions somewhere else and detached from the scripts and easily provide cmdline tools.

reox:
I have a question regarding the usage of the click library. If there is a function for doing something with default arguments set:
```
def dosomething_main(x, y=2, z=None):
    pass
```

What would be the best way to create a cli interface for it? I see, that you created a wrapper function for it like this:

```
@click.command()
@click.option("-x")
@click.option("-y", default=2)
@click.option("-z")
def dosomething(x, y, z):
    dosomething_main(x, y, z)
```

But this effectively duplicates the default values, right? But I need to have the default values in `dosomething_main`, otherwise it is hard to reuse in a python module.
How would you deal with this in your opinion?

MartinThoma:
Hm, ok, let's see:

* We have a function `dosomething_main(x, y=2, z=None)` which should definitely keep the default arguments
* We have a CLI for that function. Of course, it should have the same default arguments.

One way to do this is to externalize the defaults, e.g.:

```
defaults = {'y': 2, 'z': None}
@click.command()
@click.option("-x")
@click.option("-y", default=defaults['y'])
@click.option("-z", default=defaults['z'])
def dosomething(x, y, z):
    dosomething_main(x, y, z)

def dosomething_main(x, y=defaults['y'], z=defaults['z']):
    pass
```

Another option might be [introspection](https://stackoverflow.com/a/12627202/562769).

reox:
Thanks for the information! Probably worth a feature request ;)

reox:
Okay I send a feature request but it seems that calling the function which is decorated directly is disregarded. So it would always be better to write a extra cli function for function.

reox:
Thanks again for implementing this, it is a really nice addition! I added now a function for androdis too (not merged yet though)

Issue #423, opened 2018-02-17
readthedocs


reox:
The repo on readthedocs is owned by someone else.
Also we should setup a webhook...

reox:
@adesnos would it be possible to gain me the rights so I can setup webhooks from the androguard repository?

reox:
@adesnos ping

reox:
@adesnos @adesnosgoogle ping

reox:
I think it is easier to create a github hosted website, and a repo where the documentation is in as html...

reox:
@adesnos I have now access to the repository on rtd. Is it possible that you add the webhook or give me the rights to do it? thanks!

reox:
@adesnos @Lircyn ping

reox:
@androguard/external-contributors @adesnos ping. Anyone can help here?

PR #563, opened 2018-10-02
ENH: Add 'androguard' executable script

MartinThoma:
WIP: Only for androaxml for now

See issue #561

How this PR improves the project:

* New users will intuitively try the command `androguard`. By adding this and the use of `--help` they can discover all of the functionality of androguard.
* Moving code to the `androguard` directory makes it easier to test and possible to import in other code.
* By not removing the scripts I think this is backwards-compatible (I'm not sure how big the userbase of androguard is). I would, however, suggest to deprecate the old top level scripts.


MartinThoma:
@reox Let me know what you think. If this looks good to you, I would refactor the other commands as well. When you now install the package, you can do:

```
$ androguard --help
Usage: androguard [OPTIONS] COMMAND [ARGS]...

  Androguard is a full Python tool to play with Android files.

Options:
  --version  Show the version and exit.
  --help     Show this message and exit.

Commands:
  androaxml  Parse the AndroidManifest.xml.
```

and then 

```
$ androguard androaxml --help
Usage: androguard androaxml [OPTIONS] [FILE_]

  Parse the AndroidManifest.xml.

  Parsing is either direct or from a given APK and prints in XML format or
  saves to file.

  This tool can also be used to process any AXML encoded file, for example
  from the layout directory.

  Example:

      $ androguard androxml AndroidManifest.xml

Options:
  -i, --input TEXT   AndroidManifest.xml or APK to parse (legacy option)
  -o, --output TEXT  filename to save the decoded AndroidManifest.xml to
  --help             Show this message and exit.

```

bin/androguard
```diff
@@ -0,0 +1,68 @@
...
+@click.option(
+    '--output', '-o',
+    help='filename to save the decoded AndroidManifest.xml to',
+)
+@click.argument(
+    'file_',
+    # help='AndroidManifest.xml or APK to parse',
+    required=False,
+)
+def androaxml(input_, output, file_):
```
reox: just name sub-commands without `andro`? So `androguard axml` would be fine IMO.

bin/androguard
```diff
@@ -0,0 +1,68 @@
...
+        print("Give one file to decode!")
+        sys.exit(1)
+
+    if file_ is not None:
+        androaxml_main(file_, output)
+    elif input_ is not None:
+        androaxml_main(input_, output)
+
+
+if __name__ == '__main__':
```
reox: afaik the click doc says to use entry_points in the setup.py instead. can you do this as well?

androguard/cli.py
```diff
@@ -0,0 +1,26 @@
...
+    else:
+        print("Unknown file type")
+        return
+
+    buff = etree.tostring(axml, pretty_print=True, encoding="utf-8")
+    if outp:
+        with open(outp, "wb") as fd:
+            fd.write(buff)
+    else:
+        print(buff.decode("UTF-8"))
```
reox: click.echo should deal with bytes/str as well, right?

reox:
nice! this looks very nice! Please go for the refactoring!

androguard/cli.py
MartinThoma: Hmm... for this PR, I would like not to change this. I have never used `click.echo` and I never had problems with `print`. So if it worked before, it will still work. If it was broken before, it will still be broken.

reox: okay, no worries! Leave it as it is then. I was just reading through the click doc and they are advertising this function as the mighty saviour :smile:

bin/androguard
MartinThoma: Interesting catch! I was not aware of this. I want check how to use this in one of my projects first, then I'll adapt. I think I'll simply move `bin/androguard` to `androguard/entry_point.py`.

reox: I think having it in `bin/androguard` is fine. I think you can even have the __name__ stuff in there to be even more flexible - as long as `if __name__ ...` calls only a single function, which is also used as entry point.

subho007:
Like this PR 👍

MartinThoma:
I think I've added all commands to the `androguard` entrypoint.

Open questions for this PR:

* `androguard auto`: What is a good one-line description for this?
* `androguard odd`: Should I rename `odd` to `decompile`?

MartinThoma:
I'm not going to write tests in this PR for code that existed before and had not been tested before.

reox:
Very nice work!

> Open questions for this PR:
> 
>     * `androguard auto`: What is a good one-line description for this?

Mhh this tool is quite outdated and I don't even know how it works...
This is what the documentation says: https://androguard.readthedocs.io/en/latest/tools/androauto.html
I think it was ment to run some kind of scripts somehow...


> 
>     * `androguard odd`: Should I rename `odd` to `decompile`?

The old name was `androdd`, right? So the actual name of the tool was `dd`. But `decompile` is better IMO.
 
> 
> 
> Things open for other PRs:
> 
>     * Default arguments for functions and for CLI should always be the same: [#561 (comment)](https://github.com/androguard/androguard/issues/561#issuecomment-426590486)

According to the developers of click you should not use the same function again... But they did not say anything for default arguments. Maybe have to open a discussion in the google groups or something and ask for a good advice how to deal with such a situation.
> 
>     * Currently, a lot of functionality is in `androguard/cli/main.py`. Maybe this can be moved to other modules?

Maybe, for some parts. But we would really carefully look through the code and see which functions are useful in the modules and which are not... 

> 
>     * Testing: I have no idea how to unittest this repository. I would suggest to add `tox` (I'm not sure about nose, I've only used pytest recently)

Okay, so the current tests are in the tests folder. They are basic `unittest` tests which can be run via nose.

>     * Documentation: See [mpu](https://github.com/MartinThoma/mpu) and its [pypi page](https://pypi.org/project/mpu/0.15.0/)

Documentation, yes... So we have some here: https://androguard.readthedocs.io/en/latest/tools/tools.html
The problem is, that we can not automatically build it on readthedocs, as I can not set up webhooks. See #423 @adesnos ping
> 
>     * Style guide: I think having a style which is checked is nice. `androguard` currently gives 1643 errors with `pep8`.

hehe... yeah, there were some attempts to autopep everything but there is so much more code which I did not took any look when I fixed it for python3, so probably autopeping everything again would be a good idea.

reox:
Btw: The scripts were tested just for error outputs. So we would run all scripts with common parameters on the APKs in this repo and raise errors if the script crashed. See the travis.yml.

MartinThoma:
@reox Ok, I removed the changes to "androauto" and changed "odd" to "decompile". From my side, this PR is ready to be merged. What do you think?

MartinThoma:
Two more ideas:

* I guess you are the maintainer of androguard, right? Maybe we should add you to the `setup.py`
* If you (as the maintainer) don't know what `androauto.py` is good for ... it might be a good idea to deprecate it. (See https://github.com/MartinThoma/mpu/blob/master/mpu/decorators.py#L26 )

reox:
> 
> 
> Two more ideas:
> 
>     * I guess you are the maintainer of androguard, right? Maybe we should add you to the `setup.py`
> 

Uhm, at least I can upload it now to pypi :D So yes - maybe that would be a good idea.

>     * If you (as the maintainer) don't know what `androauto.py` is good for ... it might be a good idea to deprecate it. (See https://github.com/MartinThoma/mpu/blob/master/mpu/decorators.py#L26 )

There is plenty of stuff in androguard, where I have no clue what it does but some people seem to use it.
it looks like that androauto is older than the git repo, thus it is from the time where androguard was still on code.google.com and managed by mercurial... In principle it looks to be a useful tool. Maybe we should get more people involved into androguard who actually use it on a daily basis and can decide better what is deprecated and what is not.
For my case, I use androguard on a regular basis but only use a subset of the functions available. But it would be not wise to deprecate everything that I do not use...

MartinThoma:
I've added a help page which is similar to the other CLI tool help pages

reox:
Thank you very much for the PR!

Issue #559, opened 2018-09-25
KeyError in method get_requested_aosp_permissions


packmad:
### Describe what you wanted to do

- is this error specific to a single file? no
- if possible give a minimal working example where the error happens

```
a, d, dx = AnalyzeAPK("some.apk")
a.get_requested_aosp_permissions()

line 873, in get_requested_aosp_permissions
    if perm in list(self.permission_module["AOSP_PERMISSIONS"].keys()):
KeyError: 'AOSP_PERMISSIONS'
```

### Describe what actually happened
The dict self.permission_module is indexed by android permission name (e.g. android.permission.SEND_SMS); looks like the mapping is build in a wrong way

### System Information

- Androguard Version: 3.2.1 (latest from pip3)
- Python Version: 3.6
- Operating System: Ubuntu 18.04

reox:
Mhh I think that all permissions in the permission module are the AOSP permissions... thus, removing `["AOSP_PERMISSIONS"]` would give the correct result IMO

zqkFrank:
Hello, I meet the same problem, I try my best, but I can't solve this problem. Could you give me some suggestions? Thanks.

lukhio:
Hello. I faced the same problem. @reox is right: all the AOSP permissions are in `permission_module`, so replacing in `androguard/core/bytecodes/apk.py` at line 874:

```python
if perm in list(self.permission_module["AOSP_PERMISSIONS"].keys()):
```
by
```python
if perm in list(self.permission_module.keys()):
```
fixes this bug and indeed returns AOSP permissions:
```
$ androlyze.py foo.apk
Loading apk foo.apk...
Please be patient, this might take a while.
Androguard version 3.2.1 started
In [1]: a.get_requested_aosp_permissions()
Out[1]: 
['android.permission.ACCESS_NETWORK_STATE',
 'android.permission.READ_PHONE_STATE',
 'android.permission.INTERNET']
```

reox:
@lukhio if you like, please send in a PR!

PR #579, opened 2018-11-24
Fix KeyError in methods to get permissions

lukhio:
This is a fix for issue #559. This fixes a `KeyError` exception in `get_details_permissions`, `get_requested_aosp_permissions` and `get_requested_aosp_permissions_details`.


reox:
Thanks for fixing!

PR #583, opened 2018-11-30
Closes #582: Fix language and region utf-8 encoding

### LINKED ISSUE #582: UnicodeEncodeError: Unencoded UTF-8 unicode in ARSC._analyze
The bug is here : https://github.com/androguard/androguard/blob/a18256203d7af751c8862b04b15b15c23225b54f/androguard/core/bytecodes/axml.py#L1017

`language` and `region` are `<unicode>` type but are being used as string. This raised the following error:

```shell

UnicodeEncodeError: 'ascii' codec can't encode character u'\xa4' in position 0: ordinal not in range(128)

```
`u'\xa4'` being some unicode region name.

- Androguard Version: master
- Python Version: 2.7.10
- Operating System: Mac OS

jochasinga:
Fix unicode `language` and `region` not properly utf-8-encoded.

### DIFF
--- a/androguard/core/bytecodes/axml.py
+++ b/androguard/core/bytecodes/axml.py
@@ -1012,8 +1012,9 @@ class ARSCParser(object):
                         language = a_res_type.config.get_language()
                         region = a_res_type.config.get_country()
                         if region == "\x00\x00":
-                            locale = language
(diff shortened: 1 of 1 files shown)


reox:
thanks for findings this method :+1:

PR #587, opened 2018-12-14
Add support for APK v3 signing block

1orenz0:
Hi,

I've implemented the APK signature scheme v3 which was missing from androguard. I relied on the following sources for the implementation :

* https://source.android.com/security/apksigning/v3 : official documentation about APK V3 signing
* https://github.com/avast/apkverifier/blob/master/signingblock/schemev3.go : Go implementation of the APK V3 signing parser.

This is a pretty new scheme, so not many APK has them (it is backwards compatible with V2) but you can test it with Google Play's APK. I also took some liberty to refactor the internal code since the parsing is extremely similar between V2 and V3.

Anyway if you have remarks or questions, feel free to ask.

Cheers

### DIFF
--- a/androguard/cli/main.py
+++ b/androguard/cli/main.py
@@ -362,9 +362,10 @@ def androsign_main(args_apk, args_hash, args_all, show):
     from androguard.util import get_certificate_name_string
 
     import hashlib
+    import binascii
     import traceback
     from colorama import Fore, Style
-    from asn1crypto import x509
+    from asn1crypto import x509, keys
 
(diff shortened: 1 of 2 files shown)


reox:
Very nice! This looks very good!

reox:
Can you also use the testcases from https://android.googlesource.com/platform/tools/apksig/+/master/src/test/resources/com/android/apksig ? I saw that there are now some v3 samples as well.
The other tests are implemented here: https://github.com/androguard/androguard/blob/master/tests/test_apk.py#L137

thanks in advance!

PR #610, opened 2019-01-25
androauto.py got mistakenly removed from setup.py, readd

eighthave:
026d42777291b893238c5d63c5200dda803e123b seems to want to undo some previous
changes, but it also removed androauto.py from setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,6 +72,7 @@ setup(
     scripts=['androaxml.py',
              'androarsc.py',
              'androsign.py',
+             'androauto.py',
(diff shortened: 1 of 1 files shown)


reox:
the question is: does androauto even work? It is/was some tool to have some scripting ability... Somehow you need to put some python code into androauto to make it work?! I have no idea :(

eighthave:
I'm also OK with you removing it.  In that case, just remove it everywhere, including the script itself.  It'll always be in git history.

PR #633, opened 2019-02-07
Fix xmlparser abstraction

subho007:
@reox can you please check this PR? I have fixed few bugs as well. Abstracted the xml parsing, removed `get_element` and `get_elements` and used a better approach.

Let me know if you have any suggestion

### DIFF
--- a/androguard/core/bytecodes/apk.py
+++ b/androguard/core/bytecodes/apk.py
@@ -330,18 +330,14 @@ class APK(object):
                 self.package = self.get_attribute_value("manifest", "package")
                 self.androidversion["Code"] = self.get_attribute_value("manifest", "versionCode")
                 self.androidversion["Name"] = self.get_attribute_value("manifest", "versionName")
-                uses_permission = list(self.get_all_attribute_value("uses-permission", "name"))
-                self.permissions = self.permissions + uses_permission
+                permission = list(self.get_all_attribute_value("uses-permission", "name"))
+                self.permissions = list(set(self.permissions + permission))
 
-                maxSdkVersion = None
-                try:
-                    maxSdkVersion = int(self.get_max_sdk_version())
-                except ValueError:
-                    log.warning(self.get_max_sdk_version() + 'is not a valid value for <uses-permission> maxSdkVersion')
-                except TypeError:
-                    pass
-
-                self.uses_permissions = [[i, maxSdkVersion] for i in uses_permission]
+                for uses_permission in self.find_tags("uses-permission"):
+                    self.uses_permissions.append([
+                        self.get_value_from_tag(uses_permission, "name"),
+                        self._get_permission_maxsdk(uses_permission)
+                    ])
 
                 # getting details of the declared permissions
                 for d_perm_item in self.find_tags('permission'):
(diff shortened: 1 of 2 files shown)


subho007:
@reox fixing the testcases

androguard/core/bytecodes/apk.py
```diff
@@ -767,25 +778,6 @@ def is_multidex(self):
         dexre = re.compile("^classes(\d+)?.dex$")
         return len([instance for instance in self.get_files() if dexre.search(instance)]) > 1
 
-    def get_elements(self, tag_name, attribute, with_namespace=True):
```
reox: I would not remove them, as we saw in the past that some functions are still used...
they will still work fine, right?

androguard/core/bytecodes/apk.py
```diff
@@ -804,39 +796,63 @@ def _format_value(self, value):
...
+            tag for tag in tags if self.is_tag_matched(
+                tag, **attribute_filter
+            )
+        ]
+
+    def is_tag_matched(self, tag, **attribute_filter):
+        if len(attribute_filter) <= 0:
+            return True
+        for attr, value in attribute_filter.items():
+            _value = tag.get(attr) or tag.get(self._ns(attr))
```
reox: so this tackles the problem of no namespace, right?
Would it be a good idea to add some warning message if the tag should have a namespace but no one is found? But the question is how to know for each attribute if a namespace would be required...
The next question is what to do with files that do stuff like:
`<uses-permission android:name="some dangerous permission" name="some normal permission" />`
these would trick androguard to read the normal permission and not the dangerous one.

androguard/core/bytecodes/apk.py
```diff
@@ -804,39 +796,63 @@ def _format_value(self, value):
                     value = self.package + "." + value
         return value
 
-    def get_element(self, tag_name, attribute, **attribute_filter):
```
reox: same here, I would leave it in the code...

maybe mark both in the docstring with 
```
.. deprecated:: 3.3.5
    use :func:`whatever` instead
```

If we leave it in, we could merge it for 3.3.5, otherwise we should wait for 3.4.0 as it would break the API.

reox:
Other than the comments, i like it! Maybe have @eighthave take a look as well?

androguard/core/bytecodes/apk.py
```diff
@@ -804,39 +796,63 @@ def _format_value(self, value):
...
-
-                if skip_this_item:
-                    continue
-
-                value = item.get(self._ns(attribute))
-
-                if value is not None:
-                    return value
-        return None
+    def get_all_attribute_value(
```
reox: can you please add a docstring for the new methods?

subho007:
@reox thanks for the comments. I will be updating the PR with the suggested changes

androguard/core/bytecodes/apk.py
subho007: How would android os parse it? will it take `android:name` or `name` first? We need to test that.

androguard/core/bytecodes/apk.py
subho007: yes, this is the next thing left

androguard/core/bytecodes/apk.py
subho007: depending on that we can either place the namespace based attribute on the first half of `or` statement

subho007:
@reox I have put in a `# TODO` for now to figure out if we give `tag.get(attr)` more priority or `tag.get(self._ns(attr))` for now. The standard, for now, first try to get attribute value with a namespace, if you fail, try to get it without namespace.

subho007:
@eighthave let me know your thoughts

eighthave:
Based on a quick look, seems fine to me.  _fdroidserver_ only uses `get_element()` and only in one specific chunk of code that desparately need refactoring.

PR #634, opened 2019-02-07
redo deprecation

reox:
@subho007 sorry, I told you total BS in the last pull request... The `@DeprecationWarning` decorator is totally wrong and if you call the method you will just get an error. I'm not sure what I was thinking, but apparently I added that two years ago *shame on me*

So the correct method would be like this:
```
def deprecated_method()
    """
    .. deprecated:: 1.2.3
        use :meth:`other_method` instead. This method is likely to be removed in a later release.
    """
    warning.warn("this method is deprecated!", DeprecationWarning)
```

I fixed this now everywhere I found it.

### DIFF
--- a/androguard/core/analysis/analysis.py
+++ b/androguard/core/analysis/analysis.py
@@ -631,6 +631,11 @@ class ExternalClass:
         return self.methods.values()
 
     def GetMethod(self, name, descriptor):
+        """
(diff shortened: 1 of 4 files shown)


subho007:
hahaha @reox my bad, I should have checked it still. Thanks for the fix 😄

PR #645, opened 2019-02-18
Add Deckard to list of projects using Androguard

hrkfdn:
Deckard uses Androguard to perform static analysis an Xposed modules. It attempts to extract which hooks are placed by the module.

Thanks for your work on Androguard!


reox:
Nice tool!

hrkfdn:
Thanks! :)

PR #654, opened 2019-02-23
Drop support for legacy Python 2.7

### LINKED ISSUE #650: Drop python 2.7 support
We should consider dropping support for python 2.7, as it will be unsupported soon...

hugovk:
Fixes #650.

Adds `python_requires` to help pip.

Also upgrade syntax with https://github.com/asottile/pyupgrade and unit tests to use more useful asserts.

### DIFF
--- a/androguard/core/bytecodes/axml/__init__.py
+++ b/androguard/core/bytecodes/axml/__init__.py
@@ -101,7 +101,7 @@ def complexToFloat(xcomplex):
     return float(xcomplex & 0xFFFFFF00) * RADIX_MULTS[(xcomplex >> 4) & 3]
 
 
-class StringBlock(object):
+class StringBlock:
     """
     StringBlock is a CHUNK inside an AXML File: `ResStringPool_header`
     It contains all strings, which are used by referecing to ID's
@@ -353,7 +353,7 @@ class StringBlock(object):
                 print("{:08d} {}".format(i, repr(self.getStyle(i))))
 
 
-class AXMLParser(object):
+class AXMLParser:
     """
     AXMLParser reads through all chunks in the AXML file
     and implements a state machine to return information about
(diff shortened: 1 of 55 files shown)


androguard/core/bytecodes/axml/__init__.py
```diff
@@ -1084,12 +1084,12 @@ def _fix_value(self, value):
         if not self.__charrange or not self.__replacement:
             if sys.maxunicode == 0xFFFF:
```
reox: Do you know if there are narrow 3.x builds? otherwise we can remove this too...

reox:
Wow, this looks very nice!
To summarize what you did:
* replace all __super__(Classname, self.)... with __super__()
* remove all (object) from class definitions
* replace C string fmt with propper .format
* add regex type string if not present
* remove all unicode string definitions
* change set() to {}

There are also import from __future__ and builtins which could be removed now, as they were primary for the py2/3 compatability.
Also the one comment: This was a fix for narrow 2.x builds, but I think there are no narrow py3 builds? At least according to SO (https://stackoverflow.com/questions/14790708/do-unicode-strings-in-python-3-still-depend-on-narrow-wide-builds) this only affects up to 3.2. But as our minimal version is 3.4 (due to other issues) we should be fine.

This PR sounds like, that the next version will be 4.0.0 ;)

reox:
@androguard/external-contributors please have a look!

androguard/core/bytecodes/axml/__init__.py
hugovk: I don't know.

eighthave:
Most of the changes look simple and good.  Changing the formats could cause breakage, so I hope those cases are covered by tests.

reox:
> so I hope those cases are covered by tests.

so far, the tests work for mutf8 parsing - which was one of the most annoying things to have both working for py2 and py3.

Issue #653, opened 2019-02-22
Replace TYPE_MAP_ITEM with IntEnum in dvm_types and dvm


reox:
Replacing this with an IntEnum makes parsing a lot easier and less strings everywhere

PR #658, opened 2019-02-27
Refactor TYPE_MAP_ITEM with IntEnum

Neakxs:
Old strings are conserved in TYPE_MAP_ITEM_STRINGS for printing purposes

Resolves: #653


reox:
TYPE_MAP_ITEM_STRINGS should not be necessary. You can use .name to get the string representation of an Enum

Neakxs:
Yes you are right, I have replaced this dict with the enum property

reox:
nice work!

PR #675, opened 2019-04-18
Add 'triple' in an ast node for 'ClassInstanceCreation'.

kneep:
A class may have more than one constructor. They have different
signatures. Ast's are pure strings and numbers. Without knowing the
signatures, a parser cannot tell precisely which constructor it is
handling.


reox:
nice! thank you!

PR #679, opened 2019-04-19
Parsing somehow packed AXML files

### LINKED ISSUE #595: Is Android Namespace aware?
Some Manifest files seem to have a broken attribute name. See for example this file: https://github.com/androguard/androguard/blob/8de05b03af2b363b436cfa7646ab1bbfd4ddb13a/examples/axml/AndroidManifest_NamespaceInAttributeName.xml

they wrote the namespace string into the attribute and have the URI == -1:

```
Name 'android:versionName' starts with 'android:' prefix! The Manifest seems to be broken? Removing prefix.
Name 'android:minSdkVersion' starts with 'android:' prefix! The Manifest seems to be broken? Removing prefix.
Name 'android:targetSdkVersion' starts with 'android:' prefix! The Manifest seems to be broken? Removing prefix.
Name 'android:name' starts with 'android:' prefix! The Manifest seems to be broken? Removing prefix.
Name 'android:name' starts with 'android:' prefix! The Manifest seems to be broken? Removing prefix.
Name 'android:name' starts with 'android:' prefix! The Manifest seems to be broken? Removing prefix.
Name 'android:name' starts with 'android:' prefix! The Manifest seems to be broken? Removing prefix.
[...]
```

I thought this would not be correctly read by Android, but it does. On a Android 8 phone, all listed permissions are granted to the application on installation.
How does this work? Does the Android parser not check the attribute names?
In this case, we would need to check if the name is a compound of ns prefix and attribute name, then split it and check if there is a correct namespace mapping for this prefix.

Anyone has details how the Manifest Parsing works on android?

reox:
These files have the namespace prefix inside the string
instead of the prefix field.
If the prefix is in the namespace mapping, we will now patch it and
use it.

resolves #595


reox:
@eighthave can you test this with the fdroid server? It should fix the android: namespace stuff

eighthave:
yup fixed it!

PR #681, opened 2019-04-23
[WIP] Opcodemadness

reox:
changing and refactoring disassembly


reox:
ah yes.... I'm unsure what the best method is there... I tested my approach on several thousand dex files and found out that many have non zero values in the padding... (for Ins 10t, 20t, 30t). This looks like some sort of parser breaker...

for the others, were I changed a `H` to `BB` - what do you think would be the best method? I guess `BB` is faster but twists your brain a little bit more, because of all the LE conversion going on (but I guess thats acceptable :D)

I think I'll change them back to `BB` - thanks for reminding me!

Neakxs:
> for the others, were I changed a `H` to `BB` - what do you think would be the best method? I guess `BB` is faster but twists your brain a little bit more, because of all the LE conversion going on (but I guess thats acceptable :D)

Maybe keep the one which needs less operations after ? I don't know if such changes in formaters would affect performances

reox:
Mhh I guess using `H` and then spliiting uses a few more instructions...

subho007:
When is this massive upgrade getting completed? this looks soo promising @reox 😄

reox:
Unfortunately, the important code to parse the new opcodes of dex038 is still not done :D

reox:
Well... I found some issues with the interpretation of some opcodes and BOOM you crash the decompiler -.- So there is much much more to do unfortunately...

reox:
For some reason the build fails now but I do not see why... I have not changed anything there and on my machines the called command has an exit code == 0
@MartinThoma do you see anything obvious here? See the travis log https://travis-ci.org/androguard/androguard/jobs/599019570

maybe the problem fixes itself with the next commit, dunno...

MartinThoma:
Ok, nice that you found it :-)

Issue #684, opened 2019-04-24
Improving DEX parsing time


Neakxs:
In order to find a way to improve DEX parsing time, I used the cProfile library with a common APK.

```
Runtime : 30.15115100s

              Name               | TotTime  |  Calls   | Source

patch_string                     | 7.195434 |   539456 | mutf8.py:97    
__next__                         | 4.864838 | 16732390 | mutf8.py:81    
decode                           | 3.472627 |   539456 | mutf8.py:25    
chr                              | 2.101100 | 16192935 | mutf8.py:6     
read                             | 1.845651 |  5206598 | bytecode.py:797   
<built-in method builtins.ord>   | 1.337160 | 37593719 |
read_null_terminated_string      | 1.080101 |   102793 | dvm.py:106   
<built-in method builtins.chr>   | 0.899508 | 16192935 |
<built-in method builtins.len>   | 0.670406 | 17288235 |
get_byte                         | 0.643076 |  1021791 | dvm.py:209   
get_string                       | 0.442125 |   500248 | dvm.py:7278  
get                              | 0.417189 |   539456 | dvm.py:1865  
__init__                         | 0.373259 |    77741 | dvm.py:6715  
<built-in method builtins.isinst | 0.366974 |  5339475 |
__init__                         | 0.275484 |    76293 | dvm.py:1464  
_load_elements                   | 0.244418 |    44096 | dvm.py:3368  
<built-in method _struct.unpack> | 0.234216 |  2400031 |
__init__                         | 0.210680 |   102793 | dvm.py:1817  
readuleb128                      | 0.197332 |   566395 | dvm.py:213   
<method 'append' of 'bytearray'  | 0.183318 |  2552524 |
```

As visible, a lot's of time is spend in mutf8 package. Maybe it's possible to improve the way strings are handled during parsing.

reox:
This is an interesting observation!
Yes, maybe we can speed it up.
But if he is here
https://github.com/androguard/androguard/blob/984c0d981be2950cf0451e484f7b0d4d53bc4911/androguard/core/bytecodes/mutf8.py#L97
a lot, that means nothing good.
The mutf8 parser is needed to translate the strings found in the DEX file into a python readable form.
The Java code is much more unconstraint than python...
So if you come up with a way of not using patch_string at all, this would probably reduce the runtime a lot.

Neakxs:
In `androguard.core.bytecode.BuffHandler`, I've seen these lines :

[...]

PR #686, opened 2019-04-25
[RDY] Improving DEX parsing time

Neakxs:
I did some changes regarding the mutf8 package. I've merged `patch_string` and `decode` in one function and it seems to be faster.

**Single test with `patch_string(decode(b))` :**
```
Runtime : 30.26204100s

              Name               | TotTime  |  Calls   | Source

patch_string                     | 7.212343 |   539456 | mutf8.py:97    
__next__                         | 4.871884 | 16732390 | mutf8.py:81    
decode                           | 3.503724 |   539456 | mutf8.py:25    
chr                              | 2.161444 | 16192935 | mutf8.py:6     
read                             | 1.811831 |  5206598 | bytecode.py:797   
<built-in method builtins.ord>   | 1.305763 | 37593719 |
read_null_terminated_string      | 1.109645 |   102793 | dvm.py:106   
<built-in method builtins.chr>   | 0.933014 | 16192935 |
<built-in method builtins.len>   | 0.672361 | 17288235 |
get_byte                         | 0.648539 |  1021791 | dvm.py:209   
get_string                       | 0.442125 |   500248 | dvm.py:7278  
get                              | 0.418365 |   539456 | dvm.py:1865  
<built-in method builtins.isinst | 0.375127 |  5339475 |
__init__                         | 0.369267 |    77741 | dvm.py:6715  
__init__                         | 0.271499 |    76293 | dvm.py:1464  
_load_elements                   | 0.246015 |    44096 | dvm.py:3368  
<built-in method _struct.unpack> | 0.235059 |  2400031 |
__init__                         | 0.205666 |   102793 | dvm.py:1817  
readuleb128                      | 0.201757 |   566395 | dvm.py:213   
<method 'append' of 'bytearray'  | 0.186060 |  2552524 |
```

**Single test with `decode_and_patch(b, size)` :**
```
Runtime : 15.95257000s

              Name               | TotTime  |  Calls   | Source

decode_and_patch                 | 5.373321 |   537811 | mutf8.py:44    
read                             | 1.909715 |  5206598 | bytecode.py:795   
read_null_terminated_string      | 1.079264 |   102793 | dvm.py:97    
<built-in method builtins.chr>   | 0.816411 | 16144640 |
get_byte                         | 0.713202 |  1021791 | dvm.py:200   
get_string                       | 0.473792 |   498603 | dvm.py:7274  
<built-in method builtins.isinst | 0.377888 |  5339444 |
__init__                         | 0.303388 |    77741 | dvm.py:6714  
get                              | 0.252122 |   537811 | dvm.py:1904  
<built-in method _struct.unpack> | 0.250697 |  2400031 |
__init__                         | 0.213221 |    77741 | dvm.py:6480  
__init__                         | 0.204559 |   102793 | dvm.py:1951  
readuleb128                      | 0.200434 |   566395 | dvm.py:204   
__init__                         | 0.184938 |    82413 | dvm.py:2817  
__init__                         | 0.182839 |   100931 | dvm.py:2418  
<built-in method builtins.ord>   | 0.181255 |  5207851 |
<method 'append' of 'bytearray'  | 0.181045 |  2552524 |
add_type_item                    | 0.173937 |       35 | dvm.py:7232  
<method 'join' of 'str' objects> | 0.159373 |   561045 |
_load_elements                   | 0.142084 |    44096 | dvm.py:3371
```

I'll see if I can find something for `read` and `read_null_terminated_string` methods.

Resolves : #684

### DIFF
--- a/androguard/core/bytecode.py
+++ b/androguard/core/bytecode.py
@@ -593,60 +593,6 @@ def method2json_direct(mx):
     return json.dumps(d)
 
 
-class SV:
-
-    def __init__(self, size, buff):
-        self.__size = size
-        self.__value = unpack(self.__size, buff)[0]
-
-    def _get(self):
-        return pack(self.__size, self.__value)
-
-    def __str__(self):
-        return "0x%x" % self.__value
-
-    def __int__(self):
-        return self.__value
-
-    def get_value_buff(self):
-        return self._get()
-
-    def get_value(self):
-        return self.__value
-
-    def set_value(self, attr):
-        self.__value = attr
-
-
-class SVs:
-
-    def __init__(self, size, ntuple, buff):
-        self.__size = size
-
-        self.__value = ntuple._make(unpack(self.__size, buff))
-
-    def _get(self):
-        l = []
-        for i in self.__value._fields:
-            l.append(getattr(self.__value, i))
-        return pack(self.__size, *l)
-
-    def _export(self):
-        return [x for x in self.__value._fields]
-
-    def get_value_buff(self):
-        return self._get()
-
-    def get_value(self):
-        return self.__value
-
-    def set_value(self, attr):
-        self.__value = self.__value._replace(**attr)
-
-    def __str__(self):
-        return self.__value.__str__()
-
-
 def object_to_bytes(obj):
     """
     Convert a object to a bytearray or call get_raw() of the object
@@ -787,9 +733,6 @@ class BuffHandle:
         :param int off: starting offset
         :rtype: bytearray
         """
-        if isinstance(off, SV):
-            off = off.value
-
         return self.__buff[off:]
 
     def read(self, size):
@@ -800,9 +743,6 @@ class BuffHandle:
         :param int size: length of bytes to read
         :rtype: bytearray
         """
-        if isinstance(size, SV):
-            size = size.value
-
         buff = self.__buff[self.__idx:self.__idx + size]
         self.__idx += size
 
--- a/androguard/core/bytecodes/dvm.py
+++ b/androguard/core/bytecodes/dvm.py
@@ -1908,11 +1908,8 @@ class StringDataItem:
         string as 6 characters: \\ud853
         Valid surrogates are encoded as 32bit values, ie. \U00024f5c.
         """
-        s = mutf8.decode(self.data)
-        if len(s) != self.utf16_size:
-            raise ValueError("UTF16 Length does not match!")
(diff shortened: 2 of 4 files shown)


reox:
It is interesting, that putting both functions together makes it so much faster.

I'm not sure if we should remove the patch function completly and have a single `decode` function, which also applys the patch.

Neakxs:
> I'm not sure if we should remove the patch function completly and have a single `decode` function, which also applys the patch.

If think it's possible. The actual `decode` function is only used alone here

https://github.com/androguard/androguard/blob/984c0d981be2950cf0451e484f7b0d4d53bc4911/androguard/core/bytecodes/dvm.py#L1890-L1902

and this method doesn't seems to be used in the code.

reox:
There are are two cases: 1) get the string _as is_ from the DEX file. In this case **only**  `decode` is called. This has the advantage of getting the original string but does not work with `print()` in any case.
2) get the string in a python readable form. That's what the `patch_string` method was for.

I still do not get why calling a second function is so much slower than having it all in one function.
I know there is some overhead, but this much?
What happens if `decode_and_patch` calls `decode` and then patches the string?

Neakxs:
> I still do not get why calling a second function is so much slower than having it all in one function.
> I know there is some overhead, but this much?

I think that it's mainly due to the fact that there is a lot's of objects creation : 
- `decode` use string concatenation for each new char (`decode_and_patch` place them in an array)
- `decode` calls `chr` for each item (`decode_and_patch` keep the raw value)
- `patch_string` create a PeekIterator and then call many `ord` and `chr` (`decode_and_patch` only use `chr` when a single char is ready

> What happens if `decode_and_patch` calls `decode` and then patches the string?

I guess it will be slower because `decode` return a string.

> There are are two cases: 1) get the string _as is_ from the DEX file. In this case **only** `decode` is called. This has the advantage of getting the original string but does not work with `print()` in any case.
> 2) get the string in a python readable form. That's what the `patch_string` method was for.

I think we could add a 3rd case which is much more complicated to apply : not decoding at all during parsing and decode only when it's needed.

reox:
> I think we could add a 3rd case which is much more complicated to apply : not decoding at all during parsing and decode only when it's needed.

yes that would be an option. And in 99% of the cases, no patching is required at all.

reox:
@Neakxs should I merge this now or should it be merged with the other PR?

Neakxs:
Maybe merge it with #689 when it's ready

PR #711, opened 2019-06-03
Skip android API classes

ksg97031:
I suggest get_classes method should contain features what skip android API classes.


reox:
This is somewhat implemented in the Analysis module already via this method: https://androguard.readthedocs.io/en/latest/api/androguard.core.analysis.html#androguard.core.analysis.analysis.ClassAnalysis.is_android_api

ksg97031:
oh~ thank you

Issue #715, opened 2019-06-07
How about parse intent-filter with data element?


ksg97031:
Currently, Androguard can only parse intent-filter with action and category element.
Androguard already has the feature of parsing XML, so we can simply add [data-element](https://developer.android.com/guide/topics/manifest/data-element.html).

Therefore, how about insert parsing data element to get_intent_filters method?https://github.com/androguard/androguard/blob/79c4f750643e7a26c3b021536c1bab57dbc2e7bf/androguard/core/bytecodes/apk.py#L1130-L1162

ksg97031:
```python
         """
-        d = {"action": [], "category": []}
+        d = {"action": [], "category": [], "data": {'host':[], 'path':[], 'scheme':[]}}

         for i in self.xml:
             # TODO: this can probably be solved using a single xpath
@@ -1152,6 +1152,13 @@ class APK:
                         for ssitem in sitem.findall("category"):
                             if ssitem.get(self._ns("name")) not in d["category"]:
                                 d["category"].append(ssitem.get(self._ns("name")))
+                        for ssitem in sitem.findall("data"):
+                            if ssitem.get(self._ns("host")) not in d["data"]["host"]:
+                                d["data"]["host"].append(ssitem.get(self._ns("host")))
+                            if ssitem.get(self._ns("path")) not in d["data"]["path"]:
+                                d["data"]["path"].append(ssitem.get(self._ns("path")))
+                            if ssitem.get(self._ns("scheme")) not in d["data"]["scheme"]:
+                                d["data"]["scheme"].append(ssitem.get(self._ns("scheme")))

         if not d["action"]:
             del d["action"]
@@ -1159,6 +1166,9 @@ class APK:
         if not d["category"]:
             del d["category"]

+        if not d["data"]["host"] and not d["data"]["path"] and not d["data"]["scheme"]:
+            del d["data"]
+
```

reox:
sounds good! Do you want to create a PR?

reox:
If you can compile an example and add it to the test cases, that would also be great!

ksg97031:
Good, I'm going to create a PR for this.

PR #716, opened 2019-06-07
Code refactoring and insert data-element #715

ksg97031:
You can simply add something attribute or tag.  
https://github.com/ksg97031/androguard/blob/d5427d54f5e6d6ab805bd160401312d401ece3d8/androguard/core/bytecodes/apk.py#L1142-L1143

And this is wrong code, https://github.com/ksg97031/androguard/commit/d5427d54f5e6d6ab805bd160401312d401ece3d8#diff-7b9be541d272d4b85f40a2d14daee915L435,
`self.assertFalse(any(x != y for x, y in pairs)`) is the right code


reox:
very nice! thanks!

PR #729, opened 2019-08-30
Add the catchall handler to the end of handler list.

amimo:
The exception handler list is ordered, if the catchall handler exists, it should be the last one.


reox:
thanks for the fix!

PR #760, opened 2019-11-18
android:<id> fails getting app_icon id.

Nekmo:
Getting a app_icon fails with:

errorinvalid literal for int() with base 16: 'android:01080411'

Example apk: bbc0b4a181171c1180d9f18763e9b0e91c929fa37cadae445abe0f4d4ac673d6


reox:
okay in this case it will simply fail to get the icon, as the android namespace is not passed to the resolver. Also usually the android namespace is not available for apps.

Thanks for the PR!

PR #815, opened 2020-08-19
Add AndroBugs 2 to projects using Androguard

NDrong:


NDrong:
Repository temporarily made private.

PR #840, opened 2021-01-15
Helper tool to decompile classes/methods of a sample

cryptax:
This is a quick python helper tool. It won't thoroughly check that `--input` is a valid APK, but it does the work if you provide correct input :)


cryptax:
there is a stupid bug for class decompilation. I'm fixing it and doing a new pull request.

PR #841, opened 2021-01-15
Helper tool to decompile classes/methods [fixed]

cryptax:
Sorry, same as #840 but with silly bug fix for class decompilation!


adesnos:
Added you as a contributor so you can push that directly now !

subho007 (approved):

PR #846, opened 2021-03-11
fix python2.7 bug

TAstylex:


totoag:
Only python 3.X will be supported

PR #868, opened 2021-07-15
Fix app_name incorrect extraction

zxvzxv:
image
image
Appname will be empty string in this situation


totoag:
Seems interesting ! Could you sync with latest commit and remove chinese comments to english one ?

Issue #763, opened 2019-11-26
Jadx decompiler fails to return the source code


adamjanovsky:
### Describe what you wanted to do

When analyzing some apks, I figured that jadx is giving more suitable source-code. Thus, I wanted to use Androguard with the jadx decompiler and access the source code of classes. Sadly, it always remains empty. I followed the documentation at https://androguard.readthedocs.io/en/latest/intro/use_jadx.html. Fails on all apks. Minimal working example:

```
    a = APK('examples/android/TestsAndroguard/bin/TestActivity.apk')
    d = DalvikVMFormat(a)
    dx = Analysis(d)
    decompiler = DecompilerJADX(d, dx, jadx='/path/to/my/jadx')
    d.set_decompiler(decompiler)
    #d.set_vmanalysis(dx)

    for cls in d.get_classes():
        print(cls.source())
```

### Describe what you expected

I expected the source code to be returned/printed. upon calling `cls.source()`

### Describe what actually happened

`None` is being returned/printed.

### System Information

- Androguard Version: 3.3.5
- Python Version: 3.6
- MacOS
- Tried with old jadx 1.10 and with the newest

### Further Log Files and Output

During the decompilation, the following appears on the stdout

```
... (more lines before)
Found a class called sources/android/support/v4/content/ModernAsyncTask, which is not found by androguard!
Found a class called sources/android/support/v4/content/ContextCompat, which is not found by androguard!
Found a class called sources/android/support/v4/content/ContextCompatHoneycomb, which is not found by androguard!
Found a class called sources/android/support/v4/content/IntentCompatHoneycomb, which is not found by androguard!
... (more lines after)
Found a class called android/support/v4/view/VelocityTrackerCompat$BaseVelocityTrackerVersionImpl which is not decompiled by jadx
Found a class called android/support/v4/view/VelocityTrackerCompat$HoneycombVelocityTrackerVersionImpl which is not decompiled by jadx
... (more lines after)
```

Is my code composed correctly? Am I trying to emuleate behaviour that is not meant to be supported?

adamjanovsky:
Oh, silly me, I rushed a bit. The cause is that the paths of Androguard and Jadx don't match by a single `sources/` directory. I'll try to fix that and create PR, hopefully I manage -- never saw that code...

1eafy:
I try to use your method is not available, I use Windows OS, I choose to use regular expression to find the path.
```
        for root, dirs, files in os.walk(tmpfolder):
            for f in files:
[...]

PR #869, opened 2021-07-20
Latest JADX version places the decompiled sources in a "sources" directory ... 

tbarabosch:
... so I'm accounting for that behavior. I hope that someone can reproduce this behavior. I'm using JADX `1.2.0`.

Furthermore,  I've added the method get_sources_path to get access to the decompiled *.java files. This only works if the class `DecompilerJADX` class is instantiated with `keep_files=True`, otherwise this method returns None.

### DIFF
--- a/androguard/decompiler/decompiler.py
+++ b/androguard/decompiler/decompiler.py
@@ -634,10 +634,13 @@ class DecompilerJADX:
         self.vmx = vmx
         # Dictionary to store classnames: sourcecode
         self.classes = {}
+        self.sources_path = None
 
         # Result directory:
         # TODO need to remove the folder correctly!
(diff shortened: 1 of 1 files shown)


mastercaution:
Nice patch, but I think there is a simpler approach. Like already mentioned in https://github.com/androguard/androguard/issues/763#issuecomment-1174938958, I would suggest to only adjust the call to JADX by using the `-ds` flag instead of the `-d` flag. JADX will then place all files that would normally be written to `tmpfolder/sources` directly to `tmpfolder`.

totoag:
@mastercaution have you a PR for that ?

totoag:
seem to fixed now

PR #877, opened 2021-08-30
Ignore errors if dirty data exists in code_item

Cr1ft:
When I try to parse and remove the hardened APK, running into DalvikCode will result in an error where unpack requires a buffer of 2 bytes. This is because AndroGuard resolves all code_items according to map_list, whereas the hardened APK internals are dynamically repaired and indexed by address. May result in dirty code_item data, causing AndroGuard parsing to fail and stop. Therefore, the wrong code_item needs to be ignored

Ignore errors if dirty data exists in code_item

### DIFF
--- a/androguard/core/bytecodes/dvm.py
+++ b/androguard/core/bytecodes/dvm.py
@@ -6712,30 +6712,35 @@ class DalvikCode:
     """
 
     def __init__(self, buff, cm):
-        self.CM = cm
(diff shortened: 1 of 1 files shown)


totoag:
Ping :)

erev0s:
Given that this PR is old and the project is substantially different nowadays, I will be closing this.
Please feel free to reopen it in case it still stands.

PR #881, opened 2021-09-30
Prioritize the tags with the namespace

danielgf3:
Sample: [4c7d2912ba24e4bf1dab1175ee435f85ec223ac86967fc61a3dacbd1a97ee329](https://www.virustotal.com/gui/file/4c7d2912ba24e4bf1dab1175ee435f85ec223ac86967fc61a3dacbd1a97ee329)

```
<uses-sdk android:minSdkVersion="15" android:targetSdkVersion="23" minSdkVersion="14" targetSdkVersion="23"/>
```

For this sample, the value 14 was being returned for the **minSdkVersion** attribute, since the value without namespace had precedence.

min_sdk

### DIFF
--- a/androguard/core/bytecodes/apk.py
+++ b/androguard/core/bytecodes/apk.py
@@ -917,7 +917,7 @@ class APK:
         """
         tags = self.find_tags(tag_name, **attribute_filter)
         for tag in tags:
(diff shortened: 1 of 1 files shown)


totoag:
Fixed in https://github.com/androguard/androguard/commit/bf9ab949f88371410c9c020db108b99152f32d74

Issue #905, opened 2022-08-19
'bytes' object has no attribute 'DEX'


jacopo-degattis:
Hi everyone,
I just discovered and installed androguard and I'm trying to execute the "AnalyseAPK" function.

```python
from androguard.misc import AnalyzeAPK
a, d, dx = AnalyzeAPK("../x.apk")
```

When I executed the second line it throws this error

```
File ~/Downloads/androguard/androguard/misc.py:65, in AnalyzeAPK(_file, session, raw)
     63 dx = Analysis()
     64 for dex in a.get_all_dex():
---> 65     df = dex.DEX(dex, using_api=a.get_target_sdk_version())
     66     dx.add(df)
     67     d.append(df)

AttributeError: 'bytes' object has no attribute 'DEX'
```

I can't figure out what the problem could be, does anyone know what could it be ?

Thanks !

saucer-man:
Also meet this

SioYooo:
Run into same issue

```
2022-10-03 11:11:37.848 | INFO     | androguard.core.apk:_apk_analysis:370 - APK file was successfully validated!
2022-10-03 11:11:37.849 | DEBUG    | androguard.core.api_specific_resources:load_permissions:48 - Available API levels: 4, 5, 6, 7, 8, 9, 10, 13, 14, 15, 16, 17, 18, 19, 21, 22, 23, 24, 25, 26, 27, 28, 29
2022-10-03 11:11:37.849 | WARNING  | androguard.core.api_specific_resources:load_permissions:52 - Requested API level 31 is larger than maximum we have, returning API level 29 instead.
2022-10-03 11:11:37.849 | DEBUG    | androguard.core.api_specific_resources:load_permissions:48 - Available API levels: 4, 5, 6, 7, 8, 9, 10, 13, 14, 15, 16, 17, 18, 19, 21, 22, 23, 24, 25, 26, 27, 28, 29
2022-10-03 11:11:37.850 | DEBUG    | androguard.core.api_specific_resources:load_permissions:48 - Available API levels: 4, 5, 6, 7, 8, 9, 10, 13, 14, 15, 16, 17, 18, 19, 21, 22, 23, 24, 25, 26, 27, 28, 29
Traceback (most recent call last):
  File "/home/sioyoo/tools/BOAFuzzer/call_graph_generator.py", line 59, in <module>
    main(sys.argv)
  File "/home/sioyoo/tools/BOAFuzzer/call_graph_generator.py", line 53, in main
[...]

PR #909, opened 2022-09-01
fix dex variable override

saucer-man:
fix: https://github.com/androguard/androguard/issues/905


totoag:
already merged another PR, thx !

PR #912, opened 2022-10-15
Making DecompilerJadx Windows compatible

johnc449:
Fixing the issue addressed here https://stackoverflow.com/q/15169101/446140


totoag:
Remove jadx so no longer needed

PR #916, opened 2022-12-15
Fix AXML parsing error with manipulated

ryu-s-r:
It may fix some issue with packed or obfuscated AndroidManifest.xml example, https://www.virustotal.com/gui/file/64275082cdec19c4e9d74efb9dfcaba133d1d2312e3670ad976750c9a77752cb

### DIFF
--- a/androguard/core/axml/__init__.py
+++ b/androguard/core/axml/__init__.py
@@ -447,8 +447,9 @@ class AXMLParser:
             self._valid = False
             return
         
-        self.buff.seek(axml_header.header_size + header.size)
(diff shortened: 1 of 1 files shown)


olokos:
This causes:
`AttributeError: 'bytes' object has no attribute 'readable'` 
for me on the following line:
https://github.com/androguard/androguard/blob/master/androguard/core/axml/__init__.py#L377

This happens even with clean and properly working apk files, after trying out this change on latest code.
Any ideas? @ryu-s-r

ryu-s-r:
Oops, please check with the latest commit.

PR #917, opened 2022-12-15
FIX? Ignore? multi disk ZIP error

ryu-s-r:
Unless we have better solution for multi part zip, it would be better ignoring and pretending it is single zip file. Found that some malwares use this trick to interrupt automated analysis like below sample. https://www.virustotal.com/gui/file/2b9eb34b2d86e9119ada81f2c641086225c420d3439a3ca22e2a903ce8c2c937

### DIFF
--- a/androguard/core/apk/__init__.py
+++ b/androguard/core/apk/__init__.py
@@ -1721,9 +1721,9 @@ class APK:
 
                 # These things should not happen for APKs
                 if this_disk != 0:
-                    raise BrokenAPKError("Not sure what to do with multi disk ZIP!")
+                    log.warning("Not sure what to do with multi disk ZIP!")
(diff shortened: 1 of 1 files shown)


erev0s:
@ryu-s-r  could you please share this apk or a similar one as a poc?

ryu-s-r:
@erev0s Do you have preferred way to share this malicious apk file?

erev0s:
@ryu-s-r  feel free to use a service like https://bazaar.abuse.ch/ or you can upload it somewhere as a zip with a password and share it here.

ryu-s-r:
Here's the malware sample zipped with 'infected' password.
https://quickshare.samsungcloud.com/vx6vSvYOFX4x

PR #948, opened 2023-09-19
Fix the parsing of attributes on AXML file

Mrbenoit624:
Fix the parsing of attributes that should based on ressourceids and not Stringblock to avoid some false information due to a corrupted AndroidManifest file


hallucino:
thx for your patch !!

Issue #942, opened 2023-07-13
Installing androguard version 3.4.0a1 in Linux


mervecigdem:
Hello,

I have a question about installing androguard version 3.4.0a1 in Linux.

Previously I have downloaded it into my Windows computer and I used it without any problem. By the way thank you for this library, it helped me a lot.

Now I want to install androguard into a remote Linux server, I can access the Linux server via putty, so I can use only command line. I can only install androguard 3.3.5 there but I couldn't install 3.4.0a1.

Firstly, I tired:
pip3.9 install -U androguard[magic,GUI]
But it said current version 3.3.5 is already satisfied, so it didn't download 3.4.0a1.

Secondly, I tried:
pip3.9 install -e git+https://github.com/androguard/androguard.git#egg=androguard
I received an error: "ERROR: androguard from git+https://github.com/androguard/androguard.git#egg=androguard does not appear to be a Python project: neither 'setup.py' nor 'pyproject.toml' found."

Then, I tried:
git clone --recursive https://github.com/androguard/androguard.git
cd androguard/
pip3.9 install .
I received an error: "ERROR: Directory '.' is not installable. Neither 'setup.py' nor 'pyproject.toml' found."

So I couldn't download the new version in remote Linux server. I would appreciate it if you can suggest other download methods. Thank you so much.

erev0s:
Please try now. This should be fixed

Issue #922, opened 2023-02-25
ModuleNotFoundError: No module named 'androguard.core.bytecodes'


Drjacky:
### Describe what you wanted to do

Decompile an APK

### System Information

- Androguard Version: 3.3.5
- Python Version: 3.11.2
- Operating System: macOS

### Further Log Files and Output

```bash
.
.
.
2023-02-25 11:52:08.674 | INFO     | androguard.core.analysis.analysis:add:1435 - Adding DEX file version 37
2023-02-25 11:52:15.642 | INFO     | androguard.core.analysis.analysis:add:1458 - Added DEX in the analysis took : 0min 06s
2023-02-25 11:52:22.768 | INFO     | androguard.core.analysis.analysis:create_xref:1492 - End of creating cross references (XREF) run time: 0min 07s
2023-02-25 11:52:22.769 | INFO     | androguard.session:addAPK:160 - added APK sample.apk:c1211fdc672ac17426a311ee2c8a1d351d7514bca9d8a313e525c68d9186fbea
Traceback (most recent call last):
  File "androguard/cli.py", line 505, in <module>
    entry_point()
  File "/opt/homebrew/lib/python3.11/site-packages/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/lib/python3.11/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "androguard/cli.py", line 317, in decompile
    export_apps_to_format(fname, s, output, limit,
  File "androguard/main.py", line 84, in export_apps_to_format
    from androguard.core.bytecodes import dvm
[...]

PR #957, opened 2023-11-21
Suggestion for adding a setup.py

erev0s:
After the [big clean up](https://github.com/androguard/androguard/commit/56faaa7135966014850b3d19ce5221c3f2712eac)  there is no setup.py available which creates two main issues:
 - inconveniences when more recent versions of androguard need to be added in 3rd projects
 - people get confused on how to [install it](https://github.com/androguard/androguard/issues/942) or there are issues related with the older structure like [here](https://github.com/androguard/androguard/issues/922)

Please, let me suggest to first move the `cli.py` and `main.py` to `androguard/cli` so it has its own module and then create a basic `setup.py` where the cli is defined as an entry point.

Please feel free to edit or suggest edits.

P.S: If there is intention to proceed further with this, please lets discuss adding a pyproject.toml as well and streamline the project(CI) with a github action to push into PyPI any necessary releases.


erev0s:
Complementing to this I added the full flow to publishing into PyPI.
The published example can be found [here](https://pypi.org/project/androguard-p/) where I have bumped the version to 4.1 and as expected changed the name of the package.

Please keep in mind that if this gets accepted (which I hope it will), we will need to change the name of the project back to `androguard` in the pyproject.toml and use the correct API token for [the original PyPI androguard](https://pypi.org/project/androguard/) so it can be published on the correct project.

I am open to gladly discuss further any changes

erev0s:
@hallucino fixed contact
 - projects@erev0s.com
 - https://erev0s.com/contact/

PR #962, opened 2023-12-10
fix get_apkid and new_zip

erev0s:
based on the integration with apkInspector https://github.com/androguard/androguard/pull/956, I adjusted the get_apkid and the creation of the new zip.


eighthave:
For future reference, I'm commenting after looking at this.  Looks like it just changes `zipfile.ZipFile.open()` to `ZipEntry.parse()`

PR #1012, opened 2024-03-11
fix 'set' object is not subscriptable error.

Forgo7ten:
main_activities is of type set, but index access was used, resulting in TypeError. Use the sorted() method to convert it to a list type, and then access it through subscripts.


Forgo7ten:
@erev0s This is my first time submitting a PR. Did I mess something up?

erev0s:
hi @Forgo7ten 
thank you for your PR and thank you for your patience

PR #1015, opened 2024-03-13
function to return android:name attribute of <meta-data> entries

eighthave:
This is useful for discovering whether an app has any API Keys configured, e.g.:

https://developers.google.com/maps/documentation/android/start#get_an_android_certificate_and_the_google_maps_api_key
```xml
 <meta-data
        android:name="com.google.android.maps.v2.API_KEY"
        android:value="MY_KEY" />
```

### DIFF
--- a/androguard/core/apk/__init__.py
+++ b/androguard/core/apk/__init__.py
@@ -1115,6 +1115,14 @@ class APK:
         """
         return list(self.get_all_attribute_value("provider", "name"))
 
+    def get_meta_data_names(self, apk):
+        """
+        Return the android:name attribute of <meta-data> entries.
+
+        :rtype: a list of str
+        """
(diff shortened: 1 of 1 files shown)


erev0s:
This would not only return API keys but any meta-data element, therefore I do not believe it should be added the way it is at the moment. Additionally the `apk` in the parameter list is not required.
I will be closing this for now.

eighthave:
The function follows the pattern of functions set by `get_services()`, `get_providers()`, `get_receivers()`, etc.  I think it will be useful for multiple things, that's why i provided a generic function to fetch the names of any `<meta-data>` tag.  One clear use case is finding whether an APK has any API keys set.

I tried removing the erroneous `apk` argument, but it seems I cannot update a closed pull request.

PR #1026, opened 2024-04-01
make get_apkid() return versionCode as an integer

eighthave:
`versionCode` is strictly defined to be an integer, any other type would be an error.  So `get_apkid()` should return `versionCode` as an integer or fail trying.

### DIFF
--- a/androguard/core/apk/__init__.py
+++ b/androguard/core/apk/__init__.py
@@ -2179,9 +2179,9 @@ def get_apkid(apkfile):
                     appid = value
                 elif versionCode is None and name == 'versionCode':
                     if value.startswith('0x'):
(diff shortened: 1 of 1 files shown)


eighthave:
On second thought, since _androguard_ is primarily about analysis rather than correctness, it might make sense to keep this a str, if this field can be abused.

erev0s:
Thank you @eighthave for this PR.
Indeed `versionCode` is defined either as `int` or `long` and currently the code takes care of that with `str(int(value, 16))`. The casting to a string, I believe happens so it is aligned with other parts that return strings.
Overall though you are right, it should be treated as an integer, though given how minor this is and that it does not affect the general correctness of the output, it should be fine.

PR #1037, opened 2024-04-19
Add locale option to get_app_name() method.

Forgo7ten:
Added the locale option to the get_app_name() method. Now you can get the application name of a specific locale by adding the parameter `locale`. Instead of only getting the default application name.
Like the following example:
```python
apk.get_app_name(locale="zh-rCN")
```


Forgo7ten:
Are there any plans to develop this feature? Is there any other way I can use this feature now?
I need this feature at work.
I don't want to use the method of directly modifying the pip library. This method is more troublesome to upgrade and migrate.

erev0s:
hi @Forgo7ten , thank you for this PR and for your patience while reviewing it.

PR #1042, opened 2024-04-27
Apply type annotations

ehrenb:
Applied type annotations to much of the code-base. May want to squash commits as there are a lot of redundant messages in the log..See more detailed notes below.

- For optional values, I used the syntax `Union[type,None]` over `type|None` to keep changes compatible with Python3.9+.  `Optional[type]` could also been used here.
- In some modules, circular imports occur when a module requires another (and vice versa) for a type hint.  To avoid this error, I use `from __future__ import annotations` which postpones evaluation of annotations and still allows for hints to be type names rather than strings.
	- https://adamj.eu/tech/2021/05/15/python-type-hints-future-annotations/
	- https://adamj.eu/tech/2021/05/13/python-type-hints-how-to-fix-circular-imports/
	- https://peps.python.org/pep-0563/#abstract
- Added `TYPE_CHECKING` checks to allow future static type checkign tools to run, such as mypy


erev0s:
Thank you @ehrenb! Nicely done and well structured

Issue #1030, opened 2024-04-08
handles duplicate block IDs in APK Signing Block differently from Android/apksigner


obfusk:
If you manipulate an APK's Signing Block to have e.g. duplicate v2 Signature Blocks, Android and `apksigner` will only see the first, but `androguard` will only see the last (since it uses the ID as a key for the `_v2_blocks` dict and overwrites any previous block in that case).

```py
        # Store all blocks
        while f.tell() < end_offset - 24:
            size, key = unpack('<QI', f.read(12))
            value = f.read(size - 4)
            self._v2_blocks[key] = value
```

See https://www.openwall.com/lists/oss-security/2024/04/08/8.

obfusk:
Note that the documentation clearly states to use the first v2/v3 block:

> Locate the first APK Signature Scheme v2 Block inside the APK Signing Block.
> Locate the first APK Signature Scheme v3 Block inside the APK Signing Block.

https://source.android.com/docs/security/features/apksigning/v2#v2-verification
https://source.android.com/docs/security/features/apksigning/v3#v3-verification

obfusk:
My suggestion for now would be to simply not overwrite any existing keys (block IDs) and raise a warning if any are found:

```py
            if key in self._v2_blocks:
                logger.warning("Duplicate block ID in APK Signing Block: {}".format(key))
            else:
                self._v2_blocks[key] = value
```

For later, it might be useful to have a second API that parses and returns *all* blocks for anyone wanting to inspect (manipulated) APKs with duplicates (like my work-in-progress `apksigtool` allows).

erev0s:
Thank you once again @obfusk for pointing this out.
only the first signature block is stored now and a warn is returned for duplicate ones.

Completely agree it would be nice to store the duplicate blocks and provide also information about them so I added the label `Nice to have`. Hopefully i can pick it up sometime in the next weeks.

PR #1049, opened 2024-05-16
Sign

erev0s:
First approach towards:
https://github.com/androguard/androguard/issues/1030
https://github.com/androguard/androguard/pull/1038

### DIFF
--- a/androguard/core/apk/__init__.py
+++ b/androguard/core/apk/__init__.py
@@ -1506,6 +1506,10 @@ class APK:
         pkcs7message = self.get_file(filename)
 
         pkcs7obj = cms.ContentInfo.load(pkcs7message)
+        # TODO: should be returning the matching cert here!
+        # https://github.com/androguard/androguard/pull/1038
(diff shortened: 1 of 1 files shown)


androguard/core/apk/__init__.py
```diff
@@ -1506,6 +1506,10 @@ def get_certificate_der(self, filename:str) -> bytes:
         pkcs7message = self.get_file(filename)
 
         pkcs7obj = cms.ContentInfo.load(pkcs7message)
+        # TODO: should be returning the matching cert here!
+        # https://github.com/androguard/androguard/pull/1038
+        if len(pkcs7obj['content']['certificates']) > 1:
+            logger.warning(f"Multiple certificates found. Returning the first one!")
```
obfusk, line 1512: FYI: this f-string doesn't have any replacement fields (I'd expect a linter to catch that).

erev0s: thanks, initially i had it returning the number of certificates, but it did not add much of a value so I removed it.

PR #1050, opened 2024-05-18
V1 return correct cert if multiple

erev0s:


obfusk:
I've only looked very briefly, not done a proper review, but this code doesn't seem to handle multiple SignerInfos correctly and doesn't handle Signed Attributes at all. See:

https://android.googlesource.com/platform/tools/apksig/+/refs/tags/platform-tools-34.0.5/src/main/java/com/android/apksig/internal/apk/v1/V1SchemeVerifier.java#712

https://android.googlesource.com/platform/tools/apksig/+/refs/tags/platform-tools-34.0.5/src/main/java/com/android/apksig/internal/apk/v1/V1SchemeVerifier.java#522

obfusk:
There's should be APKs with Signed Attributes in the `apksig` test suite FYI. Though IIRC none with multiple certificates for one signer (only multiple signers).

obfusk:
Also please don't ping me in issues/PRs involving eighthave as he's left me no choice but to block him from interacting with me after [I was given no choice but to leave F-Droid](https://gitlab.com/fdroid/admin/-/issues/447).

obfusk:
This code returns the *last* certificate that has a public key that can be used to successfully verify the signature in the first SignerInfo (though it mishandles verification with Signed Attributes).

It *should* loop over either only the first (android < N) or all SignerInfos (>= N) and verify the signature using the pubkey of the *first* certificate *matching* (i.e. having the same issuer and serial number) that SignerInfo (ignoring all other certificates). And then return the certificate that was used to verify the first SignerInfo that successfully verified, if any.

> Prior to Android N, Android attempts to verify only the first SignerInfo. From N onwards, Android attempts to verify all SignerInfos and then picks the first verified SignerInfo.

https://android.googlesource.com/platform/tools/apksig/+/refs/tags/platform-tools-34.0.5/src/main/java/com/android/apksig/internal/x509/Certificate.java#65

erev0s:
@obfusk ♥ thank you so much for this detailed and constructive feedback. 

I understand and respect your request not to be involved in issues where eighthave is involved. I want to assure you that my decision to reach out to you was solely based on your exceptional experience and the quality of your work(which I stumbled upon while searching for tackling this issue). Considering the complexity of the issue and the challenges presented in both the initial PR and my own, I felt your expertise was essential for developing a robust solution. The initial PR, which had significant shortcomings, was closed due to its quality. I decided to pick up the issue myself as it sparked my interest in learning more about it. Therefore, you could consider this PR as having evolved independently of the initial one from eighthave.

Having said that, I will update it and hopefully you can take another look.

erev0s (commented):
@obfusk as you are the expert, would be great if you could take another look!

obfusk:
Happy to take another look. Just haven't had time yet :)

obfusk (commented):
I didn't check every detail. But other than the issues identified it looks pretty good to me.

androguard/core/apk/__init__.py
```diff
@@ -1499,21 +1501,153 @@ def is_androidtv(self) -> bool:
...
+            logger.info("Signing certificate referenced in SignerInfo not found in SignedData")
+        else:
+            if signer_info['signed_attrs'].native:
+                logger.info("Signed Attributes detected!")
+                signed_attrs = signer_info['signed_attrs']
+                signed_attrs_dict = OrderedDict()
+                for attr in signed_attrs:
+                    if attr['type'].dotted in signed_attrs_dict:
+                        logger.error(f"Duplicate signed attribute: {attr['type'].dotted}")
+                        return None
```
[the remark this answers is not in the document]
erev0s: Would it be more graceful to instead return None and let the use handle this however they wish?
Since there might be cases where one would wish to analyze an APK that indeed fails the verification.

androguard/core/apk/__init__.py
```diff
@@ -1499,21 +1501,153 @@ def is_androidtv(self) -> bool:
...
+                    signed_attrs_dict[attr['type'].dotted] = attr['values']
+
+                # Check content type attribute (for Android N and newer)
+                max_sdk_version = self.get_max_sdk_version()
```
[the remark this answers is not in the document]
erev0s: yes you are right the maxSdkVersion is suggested not to be set so the latest one to be used. That is why in line 1566 I have a check for it to be None.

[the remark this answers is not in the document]
obfusk: I saw the check, yeah. I'm just pointing out that *when it is set in AndroidManifest.xml* your code deviates from `apksigner` which will *never take the value from the manifest*, only allowing it to be specified on the command line.

androguard/core/apk/__init__.py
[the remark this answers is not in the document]
obfusk: Yes? I would say `get_certificate_der()` should return `None` as verification failed and there is thus no correct cert for it to return. But currently, it is `verify_signer_info_against_sig_file()` that simply returns `None`, and `get_certificate_der()` will return the wrong certificate if another call to `verify_signer_info_against_sig_file()` does return a cert.

androguard/core/apk/__init__.py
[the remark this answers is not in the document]
erev0s: a ok got your point. I followed the same flow as in the android source [here](https://android.googlesource.com/platform/tools/apksig/+/refs/tags/platform-tools-34.0.5/src/main/java/com/android/apksig/internal/apk/v1/V1SchemeVerifier.java#739).

androguard/core/apk/__init__.py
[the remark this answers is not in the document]
obfusk: There are three places in the code which currently raise a `ValueError` instead of causing `get_certificate_der()` to return `None`. I would say the same logic applies there.

androguard/core/apk/__init__.py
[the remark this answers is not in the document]
obfusk: Yeah, I'm just pointing out the value of `maxSdkVersion` in that code doesn't come from the APK (which isn't obvious from that code alone, and `minSdkVersion` by default *does* come from the APK).

androguard/core/apk/__init__.py
```diff
@@ -1501,33 +1503,200 @@ def is_androidtv(self) -> bool:
...
+
+        for cert in signed_data_certificates:
+            if cert.name == 'certificate':
+                cert_native = cert.native
+                cert_issuer = cert_native['tbs_certificate']['issuer']
+                cert_serial_number = cert_native['tbs_certificate']['serial_number']
+
+                # Create an x509.Name object for the issuer in the certificate
+                cert_issuer_name = x509.Name.build(cert_issuer)
+                cert_issuer_str = cert_issuer_name.human_friendly
```
obfusk, line 1679: The `x509.Name` is already available as `cert['tbs_certificate']['issuer']`.

Unfortunately, `.human_friendly` does not perform the same canonicalisation and will both accept and reject some certs incorrectly (e.g. the Java API does not change the order but it does normalise whitespace and case whereas this will sometimes change the order and not normalise whitespace and case).

androguard/core/apk/__init__.py
```diff
@@ -1501,33 +1503,200 @@ def is_androidtv(self) -> bool:
...
         pkcs7message = self.get_file(filename)
+        # Get the .SF
+        sf_filename = os.path.splitext(filename)[0] + '.SF'
+        sf_object = self.get_file(sf_filename)
+        # Load the signature
+        signed_data = cms.ContentInfo.load(pkcs7message)
+        # Locate the SignerInfo structure
+        signer_infos = signed_data['content']['signer_infos']
+        if not signer_infos:
+            raise ValueError('No signer information found in the PKCS7 object. The APK may not be properly signed.')
```
obfusk, line 1526: I'm not sure why this does raise an error and doesn't return `None`?

androguard/core/apk/__init__.py
```diff
@@ -1501,33 +1503,200 @@ def is_androidtv(self) -> bool:
...
+        return_certificate = None
+        list_certificates_verified = []
+        for signer_info in unverified_signer_infos_to_try:
+            try:
+                matching_certificate_verified = self.verify_signer_info_against_sig_file(
+                    signed_data,
+                    certificates,
+                    signer_info,
+                    sf_object)
+            except ValueError as e:
```
obfusk, line 1548: I'm not sure whether this should also catch a possible `TypeError` or `OSError` from oscrypto functions being called (an invalid cert seems to raise a `ValueError` as well but I'm not sure that's always the case).

Also not sure if ASN.1 parsing errors should return `None` instead.

androguard/core/apk/__init__.py
```diff
@@ -2032,7 +2201,7 @@ def get_certificates_v2(self) -> list[x509.Certificate]:
         """
         return [ x509.Certificate.load(cert) for cert in self.get_certificates_der_v2()]
 
-    def get_certificates_v1(self) -> list[x509.Certificate]:
+    def get_certificates_v1(self) -> list[Union[x509.Certificate, None]]:
```
obfusk, line 2204: This will fail for `None` values. It should append those directly and not pass them to `x509.Certificate.load()`. Or maybe they should simply be skipped here?

```py
        for x in self.get_signature_names():
            certs.append(x509.Certificate.load(self.get_certificate_der(x)))
```

obfusk: I personally prefer `Optional[x]` to `Union[x, None]`. But up to you :)

obfusk: Also: `get_certificates()` below will fail for `None` values.

androguard/core/apk/__init__.py
erev0s: hm this makes it tricky to resolve. Any ideas?

obfusk: Well, the code for that is not trivial. But we don't need to care about the specifics of the output format, we're just interested in equality. So it might be doable to just compare the Name objects with:

* multiple AVAs sorted properly
* leading zeroes in OIDs stripped
* whitespace normalised/trimmed
* strings converted to uppercase (en-US), then lowercase (en-US), then normalized using normalization form KD

I think this is the android implementation used by v1 verification:
https://android.googlesource.com/platform/libcore/+/refs/heads/android14-release/ojluni/src/main/java/sun/security/x509/AVA.java#883

obfusk:
FYI: Android/apksigner also checks for unsupported critical extensions and key usage.

And I think recent Android versions differ from `apksigner` in that they silently skip SignerInfos if the certificate cannot be found or the signed attributes are missing a message digest:

https://android.googlesource.com/platform/libcore/+/refs/heads/android14-release/ojluni/src/main/java/sun/security/pkcs/SignerInfo.java#338

androguard/core/apk/__init__.py
obfusk: FWIW it looks like directly comparing the `x509.Name` objects (not the `.native` dicts) uses the algorithm from RFC 4518, which is pretty close but not identical (it normalises whitespace and case *similarly* -- not sure if identically -- but uses NFKC instead of NFKD and doesn't sort AVAs). I've made an APK for which that fails even though `apksigner` verifies.

obfusk: Apart from being unsure about weird edge cases I found a bug in that comparison. Probably best not to use it :upside_down_face:

androguard/core/apk/__init__.py
```diff
@@ -1501,33 +1504,204 @@ def is_androidtv(self) -> bool:
...
+        """
+        matching_certificate = None
+        issuer_and_serial_number = signer_info['sid'].native
+        issuer = issuer_and_serial_number['issuer']
+        serial_number = issuer_and_serial_number['serial_number']
+
+        # Create a x509.Name object for the issuer in the SignerInfo
+        issuer_name = x509.Name.build(issuer)
+        issuer_str = issuer_name.human_friendly
+        issuer_str = unicodedata.normalize('NFKD', issuer_str.upper().lower())
```
obfusk, line 1674: Nice. This seems like a pretty neat way to handle case and unicode normalisation. I suspect there may be some edge cases but at least I can't think of a way to break it easily myself.

Unfortunately, that still leaves the fact that `.human_friendly` doesn't handle ordering correctly for our purposes. And that I can easily break to have this code produce a different result from `apksigner`/Android.

obfusk: Also, it seems `.human_friendly` does not normalise whitespace at all.

erev0s: it seems that the issuer from `CertificateChoices` and from `SignerInfo` return an ordered dictionary. Could you please provide me a dummy apk to test this against and see where it fails?

obfusk:
```sh
openssl req -x509 -newkey rsa:2048 -sha512 -outform DER -out cert-rsa-1.der -days 10000 -nodes -subj '/CN= foo+CN=bar /OU=a/OU=b/C=xx' -keyout - | openssl pkcs8 -topk8 -nocrypt -outform DER -out privkey-rsa-1.der
openssl req -x509 -newkey rsa:2048 -sha512 -outform DER -out cert-rsa-2.der -days 10000 -nodes -subj '/C=xx/CN=bar+CN=foo/OU=b/OU=a' -keyout - | openssl pkcs8 -topk8 -nocrypt -outform DER -out privkey-rsa-2.der
```

```groovy
// javacert.groovy
import java.security.cert.CertificateFactory
import java.security.cert.X509Certificate
def i = new FileInputStream(args[0])
def cert = CertificateFactory.getInstance("X.509").generateCertificate(i)
println cert.getIssuerX500Principal().getName(javax.security.auth.x500.X500Principal.CANONICAL)
```

```py
# javacert.py
import sys
from asn1crypto import x509
d = open(sys.argv[1], "rb").read()
c = x509.Certificate.load(d)
n = c["tbs_certificate"]["issuer"]
print(n.native)
print(n.human_friendly)
```

```
$ groovy javacert.groovy cert-rsa-1.der
c=xx,ou=b,ou=a,cn=bar+cn=foo
$ groovy javacert.groovy cert-rsa-2.der
ou=a,ou=b,cn=bar+cn=foo,c=xx
$ py3 javacert.py cert-rsa-1.der
OrderedDict([('common_name', [' foo', 'bar ']), ('organizational_unit_name', ['a', 'b']), ('country_name', 'xx')])
Common Name: bar ,  foo; Organizational Unit: b, a; Country: xx
$ py3 javacert.py cert-rsa-2.der
OrderedDict([('country_name', 'xx'), ('common_name', ['bar', 'foo']), ('organizational_unit_name', ['b',
'a'])])
Organizational Unit: a, b; Common Name: foo, bar; Country: xx
```

obfusk: For extra fun:

```
$ openssl req -x509 -newkey rsa:2048 -sha512 -outform DER -out cert-rsa-3.der -days 10000 -nodes -utf8 -subj '/CN=x/CN=foo  +CN=Ii   +CN=İı +CN=ẞß+CN=bar  /CN=y' -keyout - | openssl pkcs8 -topk8 -nocrypt -outform DER -out privkey-rsa-3.der
$ sed 's/bar/qux/' < cert-rsa-3.der > cert-rsa-3_.der
$ groovy javacert.groovy cert-rsa-3.der
cn=y,cn=bar+cn=foo+cn=ii+cn=ii+cn=ßss,cn=x
$ groovy javacert.groovy cert-rsa-3_.der
cn=y,cn=foo+cn=ii+cn=ii+cn=qux+cn=ßss,cn=x
$ py3 javacert.py cert-rsa-3.der
OrderedDict([('common_name', ['x', 'Ii   ', 'bar  ', 'foo  ', 'İı ', 'ẞß', 'y'])])
Common Name: y, ẞß, İı , foo  , bar  , Ii   , x
common name: y, ßss, ii , foo  , bar  , ii   , x
$ py3 javacert.py cert-rsa-3_.der
OrderedDict([('common_name', ['x', 'Ii   ', 'qux  ', 'foo  ', 'İı ', 'ẞß', 'y'])])
Common Name: y, ẞß, İı , foo  , qux  , Ii   , x
common name: y, ßss, ii , foo  , qux  , ii   , x
```

obfusk: This gets pretty close: https://gist.github.com/obfusk/b337e4f641991cc88a24d3f9567db80f

obfusk: (for comparison the abbreviations aren't needed, that was just to make it easier to compare the output)

obfusk: > This gets pretty close: https://gist.github.com/obfusk/b337e4f641991cc88a24d3f9567db80f

Improved it more. And found a difference between android and openjdk (and thus `apksigner`). But it should be really close now. I can't guaranteed there are no further edge cases, but I don't know how to break it :P

obfusk: script to generate test cert:

```sh
openssl req -x509 -quiet -newkey rsa:2048 -sha512 -outform DER -out cert-rsa-3.der -days 10000 -nodes -utf8 -config openssl.cnf -subj '/custom2=42+custom1=37+O=org/emailAddress=x@y/CN= x '$'\t\t'' 猫x/CN=foo  +CN=Ii   +CN=İı +CN=ẞß+CN=bar  +CN=zz+CN= #,;\+\\/CN=y' -keyout - | openssl pkcs8 -topk8 -nocrypt -outform DER -out privkey-rsa-3.der
```

openssl.cnf:

```
oid_section = myoids

[ myoids ]
custom1=0.1.2.3
custom2=0.1.11.3
```

androguard/core/apk/__init__.py
```diff
@@ -1501,35 +1504,256 @@ def is_androidtv(self) -> bool:
...
+            The matching certificate if found, otherwise None.
+        """
+        matching_certificate = None
+        issuer_and_serial_number = signer_info['sid'].native
+        issuer = issuer_and_serial_number['issuer']
+        serial_number = issuer_and_serial_number['serial_number']
+
+        # Create a x509.Name object for the issuer in the SignerInfo
+        issuer_name = x509.Name.build(issuer)
+        issuer_str = self.canonical_name(issuer_name)
```
obfusk, line 1672: Building a new `x509.Name` like this will break with my non-trivial examples. It's also not needed.

```suggestion
        issuer_and_serial_number = signer_info['sid'].chosen
        issuer_str = self.canonical_name(issuer_and_serial_number['issuer'])
        serial_number = issuer_and_serial_number['serial_number'].native
```

androguard/core/apk/__init__.py
```diff
@@ -1501,35 +1504,256 @@ def is_androidtv(self) -> bool:
...
+
+        for cert in signed_data_certificates:
+            if cert.name == 'certificate':
+                cert_native = cert.native
+                cert_issuer = self.canonical_name(cert.chosen['tbs_certificate']['issuer'])
+                cert_serial_number = cert_native['tbs_certificate']['serial_number']
```
obfusk, line 1678:
```suggestion
                cert_issuer = self.canonical_name(cert.chosen['tbs_certificate']['issuer'])
                cert_serial_number = cert.native['tbs_certificate']['serial_number']
```

obfusk:
> Note that unsupported critical extensions and key usage are not verified!

I *think* that apart from that it's correct now. But it's been a while so I'd have to do another full review of the final version to be sure I didn't forget or overlook anything.

There's also the matter of attribution and licensing for the canonicalisation code I wrote (since I licensed that under GPLv3+ like most of my code).

androguard/core/apk/__init__.py
```diff
@@ -2032,17 +2256,16 @@ def get_certificates_v2(self) -> list[x509.Certificate]:
         """
         return [ x509.Certificate.load(cert) for cert in self.get_certificates_der_v2()]
 
-    def get_certificates_v1(self) -> list[x509.Certificate]:
+    def get_certificates_v1(self) -> list[Union[x509.Certificate, None]]:
```
obfusk, line 2259: None values are filtered out.

androguard/core/apk/__init__.py
```diff
@@ -1501,35 +1504,256 @@ def is_androidtv(self) -> bool:
...
+                logger.info("Signed Attributes detected!")
+                signed_attrs = signer_info['signed_attrs']
+                signed_attrs_dict = OrderedDict()
+                for attr in signed_attrs:
+                    if attr['type'].dotted in signed_attrs_dict:
+                        raise ValueError(f"Duplicate signed attribute: {attr['type'].dotted}")
+                    signed_attrs_dict[attr['type'].dotted] = attr['values']
+
+                # Check content type attribute (for Android N and newer)
+                max_sdk_version = self.get_max_sdk_version()
```
obfusk, line 1580: I still think using `get_max_sdk_version()` is wrong here as `apksigner` never gets this from the APK.
Having `max_sdk_version` as an optional parameter to `get_certificate_der()` would make sense though.
Perhaps `min_sdk_version` as well, defaulting to getting it from the APK, same as `apksigner`, in that case.

erev0s:
> There's also the matter of attribution and licensing for the canonicalisation code I wrote (since I licensed that under GPLv3+ like most of my code).

ofc, I did not take that into account. Due to the code being licensed GPLv3+, it is incompatible with androguard's Apache License 2.0. Not sure what is the best angle here.

obfusk:
> ofc, I did not take that into account. Due to the code being licensed GPLv3+, it is incompatible with androguard's Apache License 2.0. Not sure what is the best angle here.

Under the circumstances I am willing to make an exception to my policy of only using strong copyleft licenses here and dual-license it so you can use it :)

erev0s:
hey @obfusk feel free to take another look and maybe i can merge it

obfusk:
Might take a little while before I can get to it. But it's on my to-do list :)

obfusk:
Still on my to-do list. But:

- I found a bug in my `canonical_name()` code I haven't had time to look into yet which I'd like to fix first.
- It seems more useful to review this after it's been modified to not use `oscrypto` as you just removed that dependency.

obfusk:
I updated the `canonical_name` gist. I could have missed more edge cases but at least I fixed the one I found. Will try to review the PR not too long after you update :)

obfusk:
I updated and moved the code to https://github.com/obfusk/x509_canonical_name.py

obfusk:
Merged without waiting for me to review?

erev0s:
Unfortunately @ping2A  merged it before discussing with me or checking the comments. @obfusk please feel free to check it even now and I can push any updates later on.

PR #1056, opened 2024-05-31
Fix invalid nsmap

danielgf3:
Namespace-altering applications detected.

## Sample:
[6a2f19dba563ca02b3a55edd74798e97db83bcc58cd77bef6a627af69facb401](https://www.virustotal.com/gui/file/6a2f19dba563ca02b3a55edd74798e97db83bcc58cd77bef6a627af69facb401/details)

```
nsmap= {'<!--': 'http://schemas.android.com/apk/res/android'}
```
```xml
<uses-permission
        <!--:name="android.permission.INTERNET" />

    <permission
        <!--:name="com.example.test.DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION"
        <!--:protectionLevel="0x2" />

    <uses-permission
        <!--:name="com.example.test.DYNAMIC_RECEIVER_NOT_EXPORTED_PERMISSION" />

    <application
        <!--:theme="@ref/0x7f100272"
        <!--:label="@ref/0x7f0f001c"
        <!--:icon="@ref/0x7f0d0000"
        <!--:allowBackup="true"
        <!--:supportsRtl="true"
        <!--:extractNativeLibs="false"
        <!--:fullBackupContent="@ref/0x7f120000"
        <!--:roundIcon="@ref/0x7f0d0001"
        <!--:appComponentFactory="androidx.core.app.CoreComponentFactory"
        <!--:dataExtractionRules="@ref/0x7f120001">
```

### DIFF
--- a/androguard/core/axml/__init__.py
+++ b/androguard/core/axml/__init__.py
@@ -988,7 +988,19 @@ class AXMLPrinter:
                         cur[-1].append(etree.Comment(comment))
 
                 logger.debug("START_TAG: {} (line={})".format(tag, self.axml.m_lineNumber))
-                elem = etree.Element(tag, nsmap=self.axml.nsmap)
+
+                try:
+                    elem = etree.Element(tag, nsmap=self.axml.nsmap)
+                except ValueError as e:
+                    logger.warning(e)
+                    # sample 6a2f19dba563ca02b3a55edd74798e97db83bcc58cd77bef6a627af69facb401
+                    # nsmap= {'<!--': 'http://schemas.android.com/apk/res/android'}
+                    if 'Invalid namespace prefix' in str(e):
+                        corrected_nsmap = self.clean_and_replace_nsmap(self.axml.nsmap, str(e).split("'")[1])
+                        elem = etree.Element(tag, nsmap=corrected_nsmap)
+                    else:
+                        raise
+
 
                 for i in range(self.axml.getAttributeCount()):
                     uri = self._print_namespace(self.axml.getAttributeNamespace(i))
@@ -1028,6 +1040,16 @@ class AXMLPrinter:
                 if len(self.axml.namespaces) > 0:
                     logger.warning("Not all namespace mappings were closed! Malformed AXML?")
                 break
+
+    def clean_and_replace_nsmap(self, nsmap, invalid_prefix):
+        correct_prefix = 'android'
+        corrected_nsmap = {}
+        for prefix, uri in nsmap.items():
+            if prefix.startswith(invalid_prefix):
+                corrected_nsmap[correct_prefix] = uri
+            else:
+                corrected_nsmap[prefix] = uri
+        return corrected_nsmap
 
     def get_buff(self) -> bytes:
         """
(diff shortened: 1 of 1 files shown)


erev0s:
Hi @danielgf3 , thank you for submitting both this and https://github.com/androguard/androguard/pull/1053.
I dont think I can download the apks from virustotal (at least not with the privileges as a normal user account).

Though for the invalid namespace i have some example APKs from malware, as this is one of the tactics employed by malware to break static analysis tools.

I will check both PRs and see what is wrong with the failed tests in the 1053.

androguard/core/axml/__init__.py
```diff
@@ -988,7 +988,18 @@ def __init__(self, raw_buff: bytes)->bytes:
...
                 logger.debug("START_TAG: {} (line={})".format(tag, self.axml.m_lineNumber))
-                elem = etree.Element(tag, nsmap=self.axml.nsmap)
+
+                try:
+                    elem = etree.Element(tag, nsmap=self.axml.nsmap)
+                except ValueError as e:
+                    logger.error(e)
+                    # nsmap= {'<!--': 'http://schemas.android.com/apk/res/android'} | pull/1056
+                    if 'Invalid namespace prefix' in str(e):
+                        corrected_nsmap = self.clean_and_replace_nsmap(self.axml.nsmap, str(e).split("'")[1])
```
obfusk, line 998: Parsing the output of an error message is very brittle as those could easily change.

Using `split("'")` to parse the output of `repr()` in such an error message even more so: e.g. it will definitely break if there's a `'` or any character that `repr()` will represent as a backslash-escape. Though it works for the given example.

Sadly, `lxml` doesn't seem to expose `libxml`'s `xmlValidateNameValue()` (which would only leave a check for `:`) or another function to easily check if a string is a valid prefix.

androguard/core/axml/__init__.py
```diff
@@ -1029,6 +1040,16 @@ def __init__(self, raw_buff: bytes)->bytes:
                     logger.warning("Not all namespace mappings were closed! Malformed AXML?")
                 break
 
+    def clean_and_replace_nsmap(self, nsmap, invalid_prefix):
+        correct_prefix = 'android'
+        corrected_nsmap = {}
+        for prefix, uri in nsmap.items():
+            if prefix.startswith(invalid_prefix):
```
obfusk, line 1047: Why `.startswith()`? AFAICT the error should return the invalid prefix. Not a prefix of the invalid prefix :)

androguard/core/axml/__init__.py
erev0s: Thank you @obfusk for taking a look at this. Indeed, I agree that error messages can easily change and this overly specific path could be irrelevant fast, though I believe it is a good approach to start with and adjust if we see things change. The logic behind it is very simple and would not require more than a few minutes to adjust it in the future.

Though its a good idea to write a test case for it, so it will be caught immediately if things changes.

erev0s: Also its a good idea when I will have enough time to further check which prefixes are allowed. I have already seen this method being used a few times.

PR #1059, opened 2024-06-01
 Invalid checksum does not throw an exception

ohyeah521:


erev0s:
Hi @ohyeah521 , rechecked this, and I believe an exception should be thrown when an invalid checksum is found.
On what grounds do you believe this should be converted to a warning?
Is there a case where a valid APK shows this behavior?

PR #1065, opened 2024-06-23
Possible integration of pyaxml [DO NOT MERGE]

Mrbenoit624:
After a discussion with Anthony Desnos I try to integrate pyaxml as AXML parser for Androguard.
For the moment I did a really partial integration with only key feature.

The objective is to integrate only when:
- All current tests of androguard is passed
- Full parsing is implemented
- A security verification is done with android implementation to see if there are no security issues

If anyone want help to do this you are welcome


erev0s:
@Mrbenoit624 @adesnos @hallucino 
Unfortunately I believe this change may not be in the best interest of the project and here are some of my key concerns:

 - AXML parsing is a core component of Androguard. By delegating this essential functionality to an external library, we lose direct control over a critical part of our toolchain. Any issues or bugs related to AXML parsing would have to be addressed by the maintainers of that library, which could lead to delays and complications in our development process. Additionally, debugging could become more complex, as we would need to navigate through another codebase to identify and resolve issues.
 - Over the years, Androguard's AXML parser has been refined to handle a wide array of edge cases. While the proposed library may cover basic functionality, our current test cases do not comprehensively address all these edge cases. Consequently, even if the library passes all existing tests, it may still fall short of the robustness we have achieved with our current implementation. This could lead to regressions and undetected issues in AXML parsing.
 - AXML parser has benefited from years of development and continuous improvement. It has proven to be reliable and effective in numerous scenarios. Replacing it with a relatively newer library would be a step backward in terms of stability and reliability. The maturity of our current parser is a significant advantage that should not be underestimated.


A quick example to illustrate this point is [this APK file](https://tria.ge/240624-wmhcbaycpl). When analyzed using the latest version of Androguard from the master branch, it works as expected. However, when the same file is processed with pyaxml, it fails to parse correctly. I mention this not to suggest that pyaxml is fundamentally flawed, but rather to highlight an edge case where our current AXML parser demonstrates its maturity and robustness.

Mrbenoit624:
Yes, thank you for your answer, I do agree with these concerns. It's what I already share to @adesnos, in addition to the benefit of having multiple implementations of AXML parser which would avoid a SPOF if vulnerabilities is identified on the parser.
I would try to improve in parrallel of androguard the implementation and thank you for your corner case I will see what go wrong in my implementation.

erev0s:
I think we can close this for now and perhaps revisit it in the future.

PR #1077, opened 2024-09-20
Convert `androguard.decompiler.dast` functions into instance or static methods

aitviid:
## What does the PR do?

Converts functions in `androguard.decompiler.dast` into instance or static methods without changing their behaviour in any way. The modifications are supposed to be cosmetic and were carefully reviewed not to affect any existing functionality. Every change was introduced in a separate commit to make reviewing easier, but feel free to squash them.

## Rationale

With the previous version of the code it was difficult to alter the functionality of the `JSONWriter` class.

For some methods in `JSONWriter`, expanding the behaviour in a subclass would require re-implementing entire code blocks or using monkey patching.

Please take a look at the following example:

In the original version of the code the `androguard.decompiler.dast` module contains a function called `method_invocation`. It is used in another function called `visit_expr`. `visit_expr` is called by `JSONWriter` in `get_cond` and `visit_switch_node` instance methods.

If a developer wanted to add extra logic for cases when `JSONWriter` detects a method invocation (e.g. to dynamically log callers), the procedure would be very complicated:

1. Use some form of monkey patching to modify the `method_invocation` and `visit_expr` functions.
2. Override `get_cond` and `visit_switch_node` methods in a way that ensures the new versions of `method_invocation` and `visit_expr` are called.

If `method_invocation` is a method inside `JSONWriter`, the procedure is much simpler - the developer only needs to override `method_invocation` in a subclass.

### DIFF
--- a/androguard/decompiler/dast.py
+++ b/androguard/decompiler/dast.py
@@ -377,42 +377,6 @@ def visit_expr(op):
     return dummy('??? Unexpected op: ' + type(op).__name__)
 
 
-def visit_ins(op, isCtor=False):
-    if isinstance(op, instruction.ReturnInstruction):
-        expr = None if op.arg is None else visit_expr(op.var_map[op.arg])
-        return return_stmt(expr)
-    elif isinstance(op, instruction.ThrowExpression):
-        return throw_stmt(visit_expr(op.var_map[op.ref]))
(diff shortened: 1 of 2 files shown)


erev0s:
nice idea and what a clean PR. Thank you

Issue #1083, opened 2024-10-17
AXML/ARSC  android14 new Feature TABLE_TYPE_FLAG_OFFSET16\ FLAG_COMPACT


FloatingGuy:
## android14 (2022)
androidfw: Add support for compact resource entries
https://github.com/aosp-mirror/platform_frameworks_base/commit/368cd19d031f5a0b219ea92531d11ccb0ee66c4d

apktool   v2.9.1  support  android 14 
https://github.com/iBotPeaches/Apktool/issues/3366
https://github.com/iBotPeaches/Apktool/pull/3206

test apk 
https://github.com/iBotPeaches/TestApks/commit/e99f9c3a29b657686fa3057f165f5ea298a9ae92#diff-daf98eb53a5485b2036ea7f583d06ea470e7ff58c65cb95eb477d2ceef4429c8R3

I have updated the axml module, adding support for the two features mentioned above, and I will push my code.

## bug traceback 
Traceback (most recent call last):
 File "/usr/local/lib/python3.9/dist-packages/androguard/run/androguard/core/apk/__init__.py", line 1225, in get_res_value
    res_parser = self.get_android_resources()
  File "/usr/local/lib/python3.9/dist-packages/androguard/run/androguard/core/apk/__init__.py", line 1658, in get_android_resources
    self.arsc["resources.arsc"] = ARSCParser(self.zip.read("resources.arsc"))
  File "/usr/local/lib/python3.9/dist-packages/androguard/run/androguard/core/axml/__init__.py", line 1517, in __init__
    ate = ARSCResTableEntry(self.buff, res_id, pc)
  File "/usr/local/lib/python3.9/dist-packages/androguard/run/androguard/core/axml/__init__.py", line 2931, in __init__
    self.item = ARSCComplex(buff, parent)
  File "/usr/local/lib/python3.9/dist-packages/androguard/run/androguard/core/axml/__init__.py", line 2989, in __init__
    self.items.append((unpack('<I', buff.read(4))[0], ARSCResStringPoolRef(buff, self.parent)))
struct.error: unpack requires a buffer of 4 bytes

erev0s:
Hi @FloatingGuy thank you for bringing it up. It was already known through https://github.com/androguard/androguard/issues/1066 and it was being worked on. The PR you [submitted](https://github.com/androguard/androguard/pull/1084) is half finished and it would require a bit more work to properly include it. Therefore I will close it and move on with https://github.com/androguard/androguard/pull/1087

[...]

PR #1084, opened 2024-10-17
Android 14 Support (Partial) 

FloatingGuy:
fixed 
- [x] 16bit entry offsets - Android 14 - support for 16-bit entry offsets  #3367
- [x]  compact entries - https://github.com/iBotPeaches/Apktool/issues/3366

reference https://github.com/iBotPeaches/Apktool/pull/3206


erev0s:
https://github.com/androguard/androguard/issues/1083#issuecomment-2453313105

Issue #1066, opened 2024-07-15
Error parsing APK name


Te-k:
### Describe what you wanted to do

Trying to parse app information of [this APK](https://www.virustotal.com/gui/file/9d28938889b08bb17564dfaef98b8b38cb9691561d0a7cdfb37a7b651db50ef9). Androguard failed to get the app name with a ARSC parsing failure.

### Describe what actually happened

```py
Importing apks/com.google.android.carrierlocation.apk
Traceback (most recent call last):
  File "/home/XXX/.pyenv/versions/3.10.4/envs/default3/lib/python3.10/site-packages/androguard/core/apk/__init__.py", line 1598, in get_android_resources
    return self.arsc["resources.arsc"]
KeyError: 'resources.arsc'


    'app_name': apk.get_app_name(),
  File "/home/XX.pyenv/versions/3.10.4/envs/default3/lib/python3.10/site-packages/androguard/core/apk/__init__.py", line 527, in get_app_name
    res_parser = self.get_android_resources()
  File "/home/XXX/.pyenv/versions/3.10.4/envs/default3/lib/python3.10/site-packages/androguard/core/apk/__init__.py", line 1604, in get_android_resources
    self.arsc["resources.arsc"] = ARSCParser(self.zip.read("resources.arsc"))
  File "/home/XXX/.pyenv/versions/3.10.4/envs/default3/lib/python3.10/site-packages/androguard/core/axml/__init__.py", line 1517, in __init__
    ate = ARSCResTableEntry(self.buff, res_id, pc)
  File "/home/XXX/.pyenv/versions/3.10.4/envs/default3/lib/python3.10/site-packages/androguard/core/axml/__init__.py", line 2931, in __init__
    self.item = ARSCComplex(buff, parent)
  File "/home/XXX/.pyenv/versions/3.10.4/envs/default3/lib/python3.10/site-packages/androguard/core/axml/__init__.py", line 2989, in __init__
    self.items.append((unpack('<I', buff.read(4))[0], ARSCResStringPoolRef(buff, self.parent)))
struct.error: unpack requires a buffer of 4 bytes
```

### System Information

- Androguard Version: master
- Python Version: 3.10.4
- Operating System: debian

erev0s:
@Te-k is this a correct apk file that is working when you install it to a device?

cryptax:
I have the same problem with `a.get_app_name()` and `a.get_android_resources()` for this APK.

This is what we have in the `AndroidManifest.xml`:

[...]

PR #1087, opened 2024-11-02
Compact resources

erev0s:
Solves https://github.com/androguard/androguard/issues/1066
Offers support for `FLAG_OFFSET16` and `FLAG_COMPACT`.

It is a continuation of a quick and dirty approach that was created in [this](https://github.com/androguard/androguard/blob/resources_flags/androguard/core/axml/__init__.py#L1546) branch


androguard/core/axml/__init__.py
```diff
@@ -1527,22 +1527,50 @@ def __init__(self, raw_buff:bytes) -> None:
...
 
                         self.packages[package_name].append(entries)
-
-                        for entry, res_id in entries:
-                            if entry != -1:
-                                ate = ARSCResTableEntry(self.buff, res_id, pc)
+                        
+                        base_offset = self.buff.tell()
+                        if base_offset != expected_entries_start:
+                            # FIXME: seems like I am missing 2 bytes here in some cases, though it does not affect the result
```
obfusk, line 1568: This occasional 2-byte difference is expected as entries are aligned to 4-byte boundaries. It should be easy to account for that instead of producing a warning.

obfusk: Yeah, IIRC your code from before should already have handled the APKs correctly. I meant you could emit the warning only if the expected start doesn't match the *position adjusted to account for alignment* here. I did mention the alignment before (and linked to my own code implementing `FLAG_OFFSET16,` `FLAG_COMPACT`, and `FLAG_SPARSE`) in my comments in 1066 :)

erev0s: I am checking this further and the misalignment seems to happen right before starting to parse the actual entries and right after having parsed the offsets to which the entries are located. As far as [I could tell](https://android.googlesource.com/platform/frameworks/base/+/refs/heads/android14-release/libs/androidfw/include/androidfw/ResourceTypes.h#1456) all cases parsing the offsets are 4-byte-aligned already, so I am missing/misunderstanding something here.

obfusk: The offsets are always 4-byte-aligned. And the header is followed by an array of offsets, followed by the entries themselves.

Before the introduction of 16-bit offsets the size of both the header and array would always be a multiple of 32 bits, meaning the end of the array would always be 4-byte-aligned and there would not be a gap between the end of the array and the 4-byte-aligned start of the entries.

But with 16-bit offsets there can be a 2-byte/16-bit gap between the (now 2-byte-aligned) end of the array and the 4-byte-aligned start of the entries because the array size is no longer always a multiple of 32 bits.

erev0s: great! thank you for explaining!

obfusk:
I didn't think this was complete yet?

PR #1090, opened 2024-11-23
updated alignment logging for resources

erev0s:


obfusk:
This failed because you only adjusted the `base_offset` for alignment when comparing it to `expected_entries_start` but still passed the unadjusted `base_offset` to `ARSCResTableEntry(...)`.

obfusk:
I would suggest:

```py
                        base_offset = self.buff.tell()
                        if base_offset + ((4 - (base_offset % 4)) % 4) != expected_entries_start:
                            logger.warning(...)
                        base_offset = expected_entries_start # <-- always use start value from header
                        for entry_offset, res_id in entries:
```

or alternatively:

```py
                        base_offset = self.buff.tell()
                        base_offset = base_offset + ((4 - (base_offset % 4)) % 4) # <-- adjust for alignment
                        if base_offset != expected_entries_start:
                            logger.warning(...)
                            base_offset = expected_entries_start # <-- use start value from header on mismatch
                        for entry_offset, res_id in entries:
```

erev0s:
thanks @obfusk, I was in a hurry and didnt notice the test failed before I merged, so i reverted as the quickest fix at that time.

PR #1096, opened 2024-12-29
updated CI tests to verify that docs build successfully

erev0s:


.github/workflows/tests.yml
```diff
@@ -31,4 +32,32 @@ jobs:
...
+        poetry config virtualenvs.create false
+
+    - name: Install Dependencies
+      run: poetry install
+
+    - name: Install Dependencies and Build Documentation
+      run: |
+        cd docs
+        pip3 install -r requirements.docs.txt
+        mkdocs gh-deploy --force
```
obfusk, line 63: @erev0s this command doesn't just build the docs but also deploys them to https://androguard.github.io/androguard (by modifying the `gh-pages` branch), which I doubt is intended *when the tests are run on any PR*.

erev0s: fair point - was thinking that different PRs would be modifying the docstrings and it would be nice to see if something messes things -- but maybe its an overkill.

PR #1136, opened 2025-08-31
Fix: normalize protectionLevel parsing in get_details_permissions

Ch0pin:
Some APKs declare permissions with protectionLevel attributes like "0x00000012" (Example `com.samsung.android.bixby.agent, sha256:7ca732b17adbe87f54a0b7fe904bc5fee825028ea88464b5bfa9c716e562de47`) . In current code these values are stored as strings, causing KeyError when looked up in protection_flags_to_attributes (which uses int keys).

This patch normalizes "0x..." strings to int and falls back to the base protection level (low 4 bits) if a combined flag is not mapped.


ping2A (approved):
Thx !!

PR #1155, opened 2025-12-22
Fix:  check worked incorrectly in case where chunk size equals chunk header size

b1dlan:
Issue occurs in `ARSCHeader` `__init__` function when it tries to parse `ResChunk_header` of `RES_XML_RESOURCE_MAP_TYPE` type chunk. At edge case when `_header_size == _size` `self._size > self.SIZE` condition becomes False so `header_ok` is False too. Parser handles it as `dummy data` and skips one byte. 

The same check done in `next` method of `ResChunkPullParser` (https://android.googlesource.com/platform/frameworks/base/+/4fafd67/tools/aapt2/ResChunkPullParser.cpp#56), but instead of strictly comparing chunk size with size of `ResChunk_header`, it compares it with `_header_size` and it allows `_size` to be equal `_header_size`

### DIFF
--- a/androguard/core/axml/__init__.py
+++ b/androguard/core/axml/__init__.py
@@ -2820,7 +2820,7 @@ class ARSCHeader:
                 == cur_pos + self._header_size + 4 + 4
             ):
                 self._size = 24
-            header_ok = self._header_size >= self.SIZE and self._size > self.SIZE
+            header_ok = self._header_size >= self.SIZE and self._size >= self._header_size
             if (self._type < RES_XML_FIRST_CHUNK_TYPE or self._type > RES_XML_LAST_CHUNK_TYPE) and header_ok:
                 break
             if cur_pos == 0 or header_ok:


erev0s:
> At edge case when `_header_size == _size` `self._size > self.SIZE` condition becomes False so `header_ok` is False too. 

but is it?

this is the current check:
```
header_ok = self._header_size >= self.SIZE and self._size > self.SIZE
```
if `self._header_size` is equal to `self._size` and `self._size` is bigger than `self.SIZE` then the `header_ok` should be `True`.

also the example you shared its an error edge case returning a null pointer - so i think i am missing something here. Could you please explain further? @b1dlan

b1dlan:
Let's consider this `network_security_config.xml` as an example:
```
xxd -e raw/res/8GD.xml
00000000: 00080003 000001d4 001c0001 000000ac   ................
00000010: 00000007 00000000 00000100 00000038   ............8...
00000020: 00000000 00000000 0000000e 0000001d   ................
00000030: 00000039 00000053 00000059 00000062   9...S...Y...b...
00000040: 61620b0b 632d6573 69666e6f 0c0c0067   ..base-config...
00000050: 74726563 63696669 73657461 63191900   certificates...c
00000060: 7261656c 74786574 66617254 50636966   leartextTrafficP
00000070: 696d7265 64657474 6e171700 6f777465   ermitted...netwo
00000080: 732d6b72 72756365 2d797469 666e6f63   rk-security-conf
00000090: 03006769 63727303 73060600 65747379   ig...src...syste
000000a0: 0d0d006d 73757274 6e612d74 726f6863   m...trust-anchor
000000b0: 00000073 00080180 00000008 00100102   s...............
000000c0: 00000024 00000002 ffffffff ffffffff   $...............
000000d0: 00000003 00140014 00000000 00000000   ................
000000e0: 00100102 00000038 00000006 ffffffff   ....8...........
000000f0: ffffffff 00000000 00140014 00000001   ................
00000100: 00000000 ffffffff 00000002 ffffffff   ................
00000110: 12000008 ffffffff 00100102 00000024   ............$...
00000120: 00000007 ffffffff ffffffff 00000006   ................
00000130: 00140014 00000000 00000000 00100102   ................
00000140: 00000038 00000008 ffffffff ffffffff   8...............
00000150: 00000001 00140014 00000001 00000000   ................
00000160: ffffffff 00000004 00000005 03000008   ................
00000170: 00000005 00100103 00000018 00000008   ................
00000180: ffffffff ffffffff 00000001 00100103   ................
00000190: 00000018 00000007 ffffffff ffffffff   ................
000001a0: 00000006 00100103 00000018 00000006   ................
000001b0: ffffffff ffffffff 00000000 00100103   ................
000001c0: 00000018 00000002 ffffffff ffffffff   ................
000001d0: 00000003                              ....
```
At `0xb4` we got chunk, which causes the problem (type `0x180` `RES_XML_RESOURCE_MAP_TYPE`, header_size `0x8` and size `0x8`) followed by `RES_XML_START_ELEMENT_TYPE (0x102)` chunk:
```
xxd -e -s 0xb4 raw/res/8GD.xml
000000b4: 00080180 00000008 00100102 00000024   ............$...
```
When `RES_XML_RESOURCE_MAP_TYPE` has `header_size == size == 0x8`, it means that this map is empty, but chunk is still valid.

Here
``` python
header_ok = self._header_size >= self.SIZE and self._size > self.SIZE
```
we got `header_ok == False` due to `self._size == self.SIZE == 0x8` that leads to skip of `dummy data`
``` python
buff.seek(cur_pos)
buff.read(1)
logger.warning(
     "Appears that dummy data are found between elements!"
)
```
at next iteration of that parser loop it reads next portion of bytes at `0xb5`
```
xxd -e -s 0xb5 raw/res/8GD.xml
000000b5: 08000801 02000000 24001001 02000000   ...........$....
```
so we got chunk of `type == 0x801` and enormous chunk size. That leads to
```
2025-12-23 12:50:40.474 | DEBUG    | androguard.core.axml:_do_next:593 - M_EVENT -1
2025-12-23 12:50:40.474 | WARNING  | androguard.core.axml:__init__:2830 - Appears that dummy data are found between elements!
2025-12-23 12:50:40.474 | DEBUG    | androguard.core.axml:_do_next:608 - NEXT HEADER <ARSCHeader idx='0x000000b4' type='2049' header_size='2048' size='33554432'>
2025-12-23 12:50:40.474 | ERROR    | androguard.core.axml:_do_next:643 - Not a XML resource chunk type: 0x0801. Skipping 33554432 bytes
2025-12-23 12:50:40.474 | ERROR    | androguard.core.axml:_do_next:610 - Error parsing resource header: Can not read over the buffer size! Offset=33554612
```
`header_ok == True` as `size > 8 bytes`, also `0x801 > RES_XML_LAST_CHUNK_TYPE` so condition will be `True`

```python
            if (self._type < RES_XML_FIRST_CHUNK_TYPE or self._type > RES_XML_LAST_CHUNK_TYPE) and header_ok:
                break
``` 
So it breaks out of `while` loop and `ARSCHeader` instantiated with `type='2049' header_size='2048' size='33554432'` and returned to `AXMLParser` `_do_next` method, which waits for XML chunk but got invalid one, so it causes:
```
2025-12-23 12:50:40.474 | ERROR    | androguard.core.axml:_do_next:643 - Not a XML resource chunk type: 0x0801. Skipping 33554432 bytes
```

Before this commit (https://github.com/androguard/androguard/pull/1134/commits/e502f53d3628163c7a8992e6d4dffa04c64ef95d) parser handled it as another portion of dummy data and just skipped those 8 bytes until it reached next chunk `0x102` due to `0x801 not in possible_types`:

```python
                if cur_pos == 0 or (
                    self._type in possible_types
                    and self._header_size >= self.SIZE
                    and self._size > self.SIZE
                ):
                    break
```

It's clear, that there are cases when chunk size is 0x8 so we can allow `_size` to **be equal** `ARSCHeader.SIZE` as it done in AOSP (https://android.googlesource.com/platform/frameworks/base/+/4fafd67/tools/aapt2/ResChunkPullParser.cpp#56). We won't skip `RES_XML_RESOURCE_MAP_TYPE` chunk and `AXMLParser` will handle this chunk correctly as an empty one as `for loop` will be skipped due to `h.size - h.header_size == 0`.

```python
            if h.type == RES_XML_RESOURCE_MAP_TYPE:
                logger.debug("AXML contains a RESOURCE MAP")
                # Check size: < 8 bytes mean that the chunk is not complete
                # Should be aligned to 4 bytes.
                if h.size < 8 or (h.size % 4) != 0:
                    logger.error(
                        "Invalid chunk size in chunk XML_RESOURCE_MAP"
                    )
                    self._valid = False
                    return

                for i in range((h.size - h.header_size) // 4):
                    self.m_resourceIDs.append(
                        unpack('<L', self.buff.read(4))[0]
                    )

                continue
```

I hope I explained it well @erev0s

erev0s:
Great thank you @b1dlan

